PR #2, opened 2020-02-29
Refactoring neo benchmark

im-kulikov:
- FastHTTP module updated to 1.9.0
- neo-go MemPool increased to 500k
- neo-go node updated to the latest revision
- sharp single node variant added for local benchmark
- neo-go module updated to v0.73.0
- rpc: refactoring and simplify RPC client
  - had rid of multiple response structs, used base response type
  - had rid of json stdlib, used easyjson module instead
  - had rid of getBlockCount and getBlock, used client methods instead
  - added GetVersion method to receive RPC Node version
  - rewritten and simplified doRPCCall method
- docker: ignored context.Canceled error, break loop instead
- added RPC error / cpu / mem summary into report
- worker: added report about RPC errors count and sent txs count
- applied changes after refactoring worker, reporter and rpc client


roman-khimov:
**neofs** benchmark?

im-kulikov:
@roman-khimov changed

PR #6, opened 2020-07-02
Update nodes to v3.0.0-preview2

fyrchik:
C# silently hangs on startup.


.docker/build/sharp.entrypoint.sh
```diff
@@ -3,11 +3,11 @@
 BIN=/neo-cli/neo-cli
 
 if [ -z "$ACC" ]; then
-  ACC=6000-privnet-blocks.acc.gz
+  ACC=dump.acc
```
roman-khimov, line 6: It's a bit strange to see it being different from the default value for neo-go entrypoint (although we always pass `ACC` explicitly).

Makefile
```diff
@@ -114,8 +114,8 @@ gen: deps
 start: stop
 ## GoSingle:
 #	## Workers:
-#	.make/runner.sh -f $(DC_GO_IR_SINGLE) -f $(DC_SINGLE) -i /dump.txs -d "GoSingle" -m wrk -w 10 -z 5m -t 30s -a node:20331
-#	make down
+	.make/runner.sh -f $(DC_GO_IR_SINGLE) -f $(DC_SINGLE) -i /dump.txs -d "GoSingle" -m wrk -w 10 -z 5m -t 30s -a node:20331
+	make down
```
roman-khimov, line 118: This shouldn't be a part of the commit.

cmd/internal/gen.go
```diff
@@ -29,43 +31,34 @@ type (
 func getWif() (*keys.WIF, error) {
 	var (
 		wifEncoded = "KxhEDBQyyEFymvfJD96q8stMbJMbZUb6D1PmXqBWZDU2WvbvVs9o"
-		version    = byte(0x00)
+		//wifEncoded = "KzfPUYDC9n2yf4fK5ro4C8KMcdeXtFuEnStycbZgX3GomiUsvX6W"
+		//wifEncoded = "KzgWE3u3EDp13XPXXuTKZxeJ3Gi8Bsm8f9ijY3ZsCKKRvZUo1Cdn"
+		//wifEncoded = "KxyjQ8eUa4FHt3Gvioyt1Wz29cTUrE4eTqX3yFSk1YFCsPL8uNsY"
+		//wifEncoded = "L2oEXKRAAMiPEZukwR5ho2S6SMeQLhcK9mF71ZnF7GvT8dU4Kkgz"
```
roman-khimov, line 37: What are these?

cmd/internal/gen.go
```diff
@@ -29,43 +31,34 @@ type (
...
-		transaction.Attribute{
-			Usage: transaction.Script,
-			Data:  fromAddressHash.BytesBE(),
-		})
+	fromAddressHash := wif.PrivateKey.GetScriptHash()
+
+	w := io.NewBufBinWriter()
+	emit.AppCallWithOperationAndArgs(w.BinWriter,
+		client.NeoContractHash, "transfer",
+		fromAddressHash, fromAddressHash, int64(1))
```
roman-khimov, line 50: `int64()` is probably not really needed.

fyrchik: Fixed.

cmd/internal/gen.go
fyrchik: Fixed.

Makefile
fyrchik: Fixed.

fyrchik:
@roman-khimov for the first test for single node it goes up to 7-8k during warmup period and then drops to 2-3k (when change views are started being initiated and block time increases to 15 seconds).
I have set txPerBlock to `40k`, it can be easily changed now.

cmd/internal/worker.go
```diff
@@ -346,7 +346,7 @@ func (d *doer) parse(ctx context.Context, startBlock int, lastTime *uint64) (las
 				continue
 			}
 
-			dt := blk.Timestamp - *lastTime
+			dt := (blk.Timestamp - *lastTime) / 1000 // Timestamp is in milliseconds.
```
roman-khimov, line 349: We're losing some precision this way, it would actually be very good to have `ms` being used everywhere now, there is a difference between 5s and 5500ms for this test.

roman-khimov:
> warmup period

How long does it take?

> change views are started being initiated and block time increases to 15 seconds

ChangeView on a single node? Wow.

cmd/internal/worker.go
fyrchik: Yes, now, we report always more than in reality. Fixed.

fyrchik:
@roman-khimov Oh, no change views, just higher block times. Forgot it was single.

PR #12, opened 2020-07-20
*: sharp node fixes

### LINKED ISSUE #3: Update to neo3-preview2
We need to branch out master-2.x and update our benchmark for Neo 3. We'll use NEO or GAS transfer transactions, initially from one address to another (to be expanded later). C# node version needs to be fixed at the moment at neo3-preview2 build (there is no protocol-level compatibility between it and current master branch of Neo).

AnnaShaleva:
Closes #3.

Now we're able to test both C# and Go implementations with both single and 4x1 configurations.

### DIFF
--- a/.docker/build/sharp.entrypoint.sh
+++ b/.docker/build/sharp.entrypoint.sh
@@ -3,7 +3,7 @@
 BIN=/neo-cli/neo-cli
 
 if [ -z "$ACC" ]; then
-  ACC=dump.acc
+  ACC=single.acc
 fi
 
 if test -f ${ACC}; then
--- a/.docker/build/sharp.entrypoint.sh
+++ b/.docker/build/sharp.entrypoint.sh
@@ -6,8 +6,8 @@ if [ -z "$ACC" ]; then
   ACC=single.acc
 fi
 
-if test -f ${ACC}; then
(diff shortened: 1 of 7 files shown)


.docker/build/sharp.entrypoint.sh
```diff
@@ -6,8 +6,8 @@ if [ -z "$ACC" ]; then
   ACC=single.acc
 fi
 
-if test -f ${ACC}; then
-    cp /${ACC} /neo-cli/chain.acc
+if test -f /$ACC; then
```
roman-khimov, line 9: `/"$ACC"`, protect var expansion please, if ACC has spaces for any reason you'd be in trouble.

AnnaShaleva:
Actually, I found one more issue: Go node has outdated `sendrawtransaction` RPC response form. It should return transaction hash instead of `bool` value in case of success. I managed this issue locally by changing `(c *RPCClient) SendTX` response variable from `bool` to
```
var res struct {
	Hash util.Uint256 `json:"hash"`
}
```
for proper C# node testing.

.docker/build/sharp.entrypoint.sh
AnnaShaleva: Also changed go shell script.

PR #14, opened 2020-08-12
*: update to the preview3

fyrchik:
Update both Go and C# to preview3 version.


.docker/build/Dockerfile.sharp
```diff
@@ -44,7 +46,7 @@ RUN wget -O /opt/neo-cli.zip ${URL} && \
     cp /temp/neo-cli/Microsoft.AspNetCore.ResponseCompression.dll /neo-cli && \
```
roman-khimov, line 46: It shouldn't be needed now.

.docker/build/Dockerfile.golang
```diff
@@ -6,7 +6,7 @@ WORKDIR /neo-go
 # Install deps:
 RUN apk add --no-cache git
 
-ARG REV="v0.90.0"
+ARG REV="master"
```
im-kulikov, line 9: bad idea because makes it impossible to reproduce at different times

fyrchik: Yeah, I will change it right after `0.91.0` is out.

roman-khimov: I think we want to set it to some recent commit for now (to merge it correctly) and then of course change to 0.91.0 when it's out.

.docker/build/Dockerfile.sharp
roman-khimov: But it's still there, any reason for that?

.docker/ir/go.protocol.privnet.one.yml
```diff
@@ -1,8 +1,9 @@
 ProtocolConfiguration:
   Magic: 56753
-  MemPoolSize: 5000
+  MemPoolSize: 500000
```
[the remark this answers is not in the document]
roman-khimov: As this one too.

.docker/build/Dockerfile.sharp
```diff
@@ -44,7 +46,7 @@ RUN wget -O /opt/neo-cli.zip ${URL} && \
...
     rm -r /temp
 
-ENV MODULES="RpcServer SystemLog"
+ENV MODULES="RpcServer"
```
roman-khimov, line 49: And BTW, this probably requires some changes to C# node configuration (see default configs there).

fyrchik: Fixed.

.docker/build/Dockerfile.golang
fyrchik: Fixed.

PR #15, opened 2020-08-14
cmd: change SendTx expected result

AnnaShaleva:
RPC API has been changed, `sendrawtransaction` returns tx hash now.

### DIFF
--- a/cmd/internal/client.go
+++ b/cmd/internal/client.go
@@ -17,6 +17,7 @@ import (
 	"github.com/nspcc-dev/neo-go/pkg/config/netmode"
 	"github.com/nspcc-dev/neo-go/pkg/core/block"
 	"github.com/nspcc-dev/neo-go/pkg/io"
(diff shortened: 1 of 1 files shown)


cmd/internal/client.go
```diff
@@ -141,12 +142,14 @@ func (c *RPCClient) GetVersion(ctx context.Context) (string, error) {
...
-	res := false
+	var res struct {
+		Hash util.Uint256 `json:"hash"`
+	}
 	rpc := fmt.Sprintf(`{"jsonrpc": "2.0", "id": 1, "method": "sendrawtransaction", "params": ["%s"]}`, tx)
 
 	if err := c.doRPCCall(ctx, rpc, &res); err != nil {
 		return err
-	} else if !res {
+	} else if res.Hash.Equals(util.Uint256{}) {
```
roman-khimov, line 152: It's be better to compare it with a real transaction hash, but I guess we just don't have it there.

PR #18, opened 2020-08-19
util: add plotting python script

AnnaShaleva:


roman-khimov (approved):
Well, if it really does produce some pictures, so be it, it's a good enough starter.

PR #21, opened 2020-08-24
cmd: be carefull with counting zero TPS values

AnnaShaleva:
1. Refactor some old 2-x code.
2. Do not pay attention to zero TPS values at the start and at the end of the test.


cmd/internal/worker.go
```diff
@@ -29,6 +29,7 @@ type (
 		waiter       *sync.WaitGroup
 		countTxs     *atomic.Int32 // stores count of completed queries
 		countErr     *atomic.Int32
+		isStart      *atomic.Bool
```
roman-khimov, line 32: `hasStarted`?

roman-khimov (approved):

PR #28, opened 2020-08-28
*: C# RocksDBStore, bench concurrent HTTP connections, Go monitoring and others 

AnnaShaleva:
A set of small but important fixes.

### DIFF
--- a/cmd/internal/client.go
+++ b/cmd/internal/client.go
@@ -77,7 +77,7 @@ func (v errorResponse) Error() string {
 }
 
 // NewRPCClient creates new client for RPC communications.
-func NewRPCClient(v *viper.Viper) *RPCClient {
+func NewRPCClient(v *viper.Viper, maxConnsPerHost int) *RPCClient {
 	var addresses []string
 	for _, addr := range v.GetStringSlice("rpcAddress") {
 		if addr == "" {
@@ -108,7 +108,7 @@ func NewRPCClient(v *viper.Viper) *RPCClient {
 		MaxIdemponentCallAttempts: 1, // don't repeat queries
 		ReadTimeout:               timeout,
 		WriteTimeout:              timeout,
-		MaxConnsPerHost:           5_000,
+		MaxConnsPerHost:           maxConnsPerHost,
 	}
 
 	return &RPCClient{
--- a/cmd/internal/client.go
+++ b/cmd/internal/client.go
@@ -57,7 +57,13 @@ type (
 		addr []string
 		len  int32
 		inc  *atomic.Int32
-		cli  *fasthttp.Client
+		// The only txSender's duty is to send `sendrawtransaction` requests in
+		// order not to affect bench results by sending service requests via the
+		// same connection. txSender has different fasthttp settings than blockRequester.
+		txSender *fasthttp.Client
+		// blockRequester should do the rest of work, e.g. fetch blocks count, fetch
+		// blocks and etc.
+		blockRequester *fasthttp.Client
 
 		timeout time.Duration
 	}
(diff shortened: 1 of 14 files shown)


cmd/internal/client.go
```diff
@@ -108,7 +108,13 @@ func NewRPCClient(v *viper.Viper) *RPCClient {
 		MaxIdemponentCallAttempts: 1, // don't repeat queries
 		ReadTimeout:               timeout,
 		WriteTimeout:              timeout,
-		MaxConnsPerHost:           5_000,
+	}
+
+	switch mode {
+	case ModeWorker:
+		cli.MaxConnsPerHost = 5_000
```
roman-khimov: Wait a minute, but that's not correct also. We're supposed to have exactly the number of connections specified.

cmd/internal/client.go
```diff
@@ -186,10 +200,10 @@ func (c *RPCClient) GetBlock(ctx context.Context, index int) (*block.Block, erro
 func (c *RPCClient) GetBlockCount(ctx context.Context) (int, error) {
 	num := 0
 	rpc := fmt.Sprintf(`{"jsonrpc": "2.0", "id": 1, "method": "getblockcount", "params": []}`)
-	return num, c.doRPCCall(ctx, rpc, &num)
+	return num, c.doRPCCall(ctx, rpc, &num, false)
 }
 
-func (c *RPCClient) doRPCCall(_ context.Context, call string, result interface{}) error {
+func (c *RPCClient) doRPCCall(_ context.Context, call string, result interface{}, isSendRawTransaction bool) error {
```
roman-khimov, line 206: Maybe `doRPCCall(..., client fasthttp.Client)`? Looking at `c.doRPCCall(ctx, rpc, &num, false)` it's very hard to guess what this `false` is all about.

cmd/internal/client.go
```diff
@@ -104,24 +110,32 @@ func NewRPCClient(v *viper.Viper, mode BenchMode) *RPCClient {
 		timeout = v
 	}
 
-	cli := &fasthttp.Client{
+	cliSender := &fasthttp.Client{
```
roman-khimov, line 113: `txSender`?

cmd/internal/client.go
```diff
@@ -104,24 +110,32 @@ func NewRPCClient(v *viper.Viper, mode BenchMode) *RPCClient {
...
 	switch mode {
 	case ModeWorker:
-		cli.MaxConnsPerHost = 5_000
+		cliSender.MaxConnsPerHost = 5_000
 	case ModeRate:
-		cli.MaxConnsPerHost = 1
+		cliSender.MaxConnsPerHost = 1
+	}
+
+	cliParser := &fasthttp.Client{
```
roman-khimov, line 126: `blockRequester`?

cmd/internal/worker.go
```diff
@@ -231,6 +231,10 @@ loop:
 
 			if err := d.cli.SendTX(ctx, d.dump.Transactions[i]); err != nil {
 				d.countErr.Inc()
+				// we should wait a little, otherwise we just throw tens of transactions
+				// away during the single transaction is being processed and the test-run
+				// ends very early
+				time.Sleep(time.Millisecond)
```
roman-khimov, line 237: Something's hairy here also. We're supposed to be sending transactions synchronously in this mode, so you send a transaction, wait for result, then send another one, etc. This way errors or magic timeouts shouldn't affect the result in any way. But this change suggests that it's just not the case.

cmd/gen/main.go
```diff
@@ -10,7 +10,7 @@ import (
 var (
 	inp = flag.String("inp", "", "Path to read dump transactions.")
 	out = flag.String("out", "./dump.txs", "Path to dump transactions.")
-	cnt = flag.Int("cnt", 1_000_000, "Count of txs that would be generated.")
+	cnt = flag.Int("cnt", 2_000_000, "Count of txs that would be generated.")
```
roman-khimov, line 13: 1000000/1000/60 = 16m40s and we're benchmarking for 5m. So why do we need it?

roman-khimov (changes requested):

roman-khimov (approved):

PR #34, opened 2020-09-14
*: configuration improvements

AnnaShaleva:
1. Generate nodes configurations
2. Remove unnecessary SharpSingle+GoRPC configuration
3. Add Make targets for standard test runs


.docker/ir/sharp.protocol.template.yml
```diff
@@ -0,0 +1,54 @@
...
+    StandbyCommittee:
+      #@ for j in range(1,len(nodes_info)):
+      - #@ nodes_info[j][3]
+      #@ end
+    SeedList:
+      #@ for j in range(1,len(nodes_info)):
+      - #@ "node_{}:{}".format(nodes_info[j][0], nodes_info[j][1])
+        #@ end
+    #@ end
+#@ end
```
roman-khimov, line 54: Please add newline

Makefile
```diff
@@ -113,19 +113,26 @@ gen: deps
...
 		&& cd cmd/ \
 		&& go run ./dump -out ../$(BUILD_DIR)/dump.acc
 
+# Generate configurations for single-node and four-nodes networks from templates
+config: deps
+	@echo "=> Generate configurations for single-node and four-nodes networks from templates"
+	@set -x \
+		&& cd ./cmd \
+		&& ../ytt -f ../.docker/ir/sharp.protocol.template.yml -o json | go run ./config/ --node sharp \
+		&& ../ytt -f ../.docker/ir/go.protocol.template.yml | go run ./config/ --node go
```
roman-khimov, line 143: I think it's better integrate this into `config` tool rather than installing `ytt`.

.gitignore
```diff
@@ -4,4 +4,7 @@ vendor
 .docker/ir/out/*
 !.docker/ir/out/.gitkeep
 *.txs
-*.log
\ No newline at end of file
+*.log
+.docker/ir/go.protocol.privnet.*.yml
+.docker/ir/sharp.protocol*.json
+.docker/ir/sharp.config.*.json
```
roman-khimov, line 10: Newline>

README.md
```diff
@@ -115,6 +115,7 @@ TPS
   Targets:
 
     build    Build all images
+    config   Generate configurations for single-node and four-nodes networks from scaffolds
```
roman-khimov: from templates.

Makefile
```diff
@@ -132,107 +142,195 @@ config: deps
...
+	.make/runner.sh -f $(DC_SHARP_IR) -f $(DC_GO_RPC) -i /dump.txs -d "Sharp4x_GoRPC" -m rate -q 60 -z 5m -t 30s -a go-node:20331
+	make stop
+
+start.SharpFourNodesGoRPC300rate: prepare
+	.make/runner.sh -f $(DC_SHARP_IR) -f $(DC_GO_RPC) -i /dump.txs -d "Sharp4x_GoRPC" -m rate -q 300 -z 5m -t 30s -a go-node:20331
+	make stop
+
+start.SharpFourNodesGoRPC1000rate: prepare
+	.make/runner.sh -f $(DC_SHARP_IR) -f $(DC_GO_RPC) -i /dump.txs -d "Sharp4x_GoRPC" -m rate -q 1000 -z 5m -t 30s -a go-node:20331
+	make stop
```
roman-khimov, line 336: Newline.

roman-khimov (changes requested):

Makefile
```diff
@@ -115,20 +115,30 @@ pull:
...
+
+# IGNORE: create transactions dump
+dump.txs: cmd/gen/main.go
 	@echo "=> Generate transactions dump"
 	@set -x \
 		&& cd cmd/ \
 		&& go run ./gen -out ../dump.txs
 
-dump.single: deps config
+dump.single: deps config single.acc
```
roman-khimov, line 128: `.docker/build/single.acc`?

Makefile
```diff
@@ -115,20 +115,30 @@ pull:
...
+# IGNORE: generate block dump for the single-node network
+single.acc: cmd/dump/main.go cmd/dump/chain.go
 	@echo "=> Generate block dump for the single node network"
 	@set -x \
 		&& cd cmd/ \
 		&& go run ./dump -single -out ../$(BUILD_DIR)/single.acc
 
 # Generate `dump.acc` for the 4-node network
-dump: deps config
+dump: deps config dump.acc
```
roman-khimov, line 138: `.docker/build/dump.acc`?

roman-khimov (commented):
It's not that bad actually (ytt), some duplication is annoying but that still is better than what we have now.

Makefile
AnnaShaleva: It also depends on `cmd/dump/main.go` and `cmd/dump/chain.go` as we'd like to re-generate dumps after changing these two files. So as a result we have the following:
```
dump.single: deps config .docker/build/single.acc cmd/dump/main.go cmd/dump/chain.go
```
Is it OK?

AnnaShaleva:
> It's not that bad actually (ytt), some duplication is annoying but that still is better than what we have now.

I also placed the common C# and Go configurations into special separate data file in order to avoid code duplication.

roman-khimov (approved):

PR #39, opened 2020-09-24
cmd, plot: plot milliseconds per block

AnnaShaleva:


cmd/bench/main.go
```diff
@@ -128,6 +128,22 @@ func main() {
 		log.Fatalf("could not fetch last block: %v", err)
 	}
 
+	log.Println("Waiting for an empty block to be processed")
+	startBlockIndex := blk.Index
+	for attempt := 0; attempt < 30; attempt++ {
```
roman-khimov, line 133: Can't we make it depending on configured per-block time?

cmd/bench/main.go
```diff
@@ -128,6 +128,22 @@ func main() {
...
+		blk, err = client.GetLastBlock(ctx)
+		if err != nil {
+			log.Fatalf("could not fetch last block: %v", err)
+		}
+		if blk.Index > startBlockIndex {
+			break
+		}
+		time.Sleep(time.Second)
```
roman-khimov, line 141: Half a second?

cmd/bench/main.go
```diff
@@ -128,6 +128,22 @@ func main() {
...
+	}
+	if blk.Index == startBlockIndex {
+		log.Fatalf("Blocks were not processed for a long time since test start")
```
roman-khimov, line 144: `timeout waiting for a new empty block`

plot.py
```diff
@@ -2,54 +2,54 @@
 import os
 import matplotlib.pyplot as plt
 
-# files batch is a dictionary {plot_name:[[log_file, legend_string, line colour],...]}
+# files batch is a dictionary {plot_name:(DefaultMillisecondsPerBlock, [[log_file, legend_string, line colour],...])}
```
roman-khimov, line 5: It duplicates node's configuration. Can we avoid that?

roman-khimov (changes requested):

cmd/bench/main.go
AnnaShaleva: Are we OK with 10 attempts and an interval = defaultMSPerBlock?

cmd/bench/main.go
AnnaShaleva: DefaultMSPerBlock?

plot.py
AnnaShaleva: Bench and nodes configurations are split apart and bench isn't aware of the network configuration (single or multiple nodes). I tried to pass MSPerBlock as an argument to the runner.sh, but it was quite painful to pars configurations to get MSPerBlock. 

In the end I came to the other solution: just mount the corresponding configuration for the bench.

cmd/bench/main.go
```diff
@@ -137,6 +137,22 @@ func main() {
...
+	startBlockIndex := blk.Index
+	for attempt := 0; attempt < 10; attempt++ {
+		blk, err = client.GetLastBlock(ctx)
+		if err != nil {
+			log.Fatalf("could not fetch last block: %v", err)
+		}
+		if blk.Index > startBlockIndex {
+			break
+		}
+		time.Sleep(time.Duration(msPerBlock) * time.Millisecond)
```
roman-khimov, line 150: What I wanted to get here by making it less than `msPerBlock` is smoother test start. The first block is a bit special in that we don't push transactions all the time between it and the previous one, so we either need to throw it away from results or try to minimize the time frame between the previous block and worker threads start. Making it `/2` or even `/4` would help with this.

And waiting for more than 4*msPerBlock is probably a waste of time, either the nodes produce blocks or they don't. Initially there is no load on the system, so should be creating them on time.

roman-khimov (changes requested):

plot.py
```diff
@@ -23,31 +23,31 @@
...
-    '4 nodes 60 rate': [["Go4x1_rate_60.log", "go", 'blueviolet'],
+    '4 nodes 60 rate':[["Go4x1_rate_60.log", "go", 'blueviolet'],
                           ["no_VT_Go4x1_rate_60.log", "go, no VT", 'orangered'],
                           ["Sharp4x_GoRPC_rate_60.log", "c# + go RPC", 'aqua']],
-    '4 nodes 300 rate': [["Go4x1_rate_300.log", "go", 'blueviolet'],
+    '4 nodes 300 rate':[["Go4x1_rate_300.log", "go", 'blueviolet'],
                             ["no_VT_Go4x1_rate_300.log", "go, no VT", 'orangered'],
                             ["Sharp4x_GoRPC_rate_300.log", "c# + go RPC", 'aqua']],
-    '4 nodes 1000 rate': [["Go4x1_rate_1000.log", "go", 'blueviolet'],
+    '4 nodes 1000 rate':[["Go4x1_rate_1000.log", "go", 'blueviolet'],
```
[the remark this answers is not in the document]
AnnaShaleva: Yes, I just removed them.

roman-khimov (approved):

PR #40, opened 2020-10-01
config: explicitly set MaxGasInvoe RPC configuration to -1

AnnaShaleva:
We should be able to use bench environment for manual tests, so let's
not limit MaxGasInvoke, otherwise it's impossible to use
`invokefunction` RPC call.

### DIFF
--- a/.docker/ir/go.protocol.template.yml
+++ b/.docker/ir/go.protocol.template.yml
@@ -62,6 +62,7 @@ ApplicationConfiguration:
     Enabled: true
     EnableCORSWorkaround: false
     Port: #@ data.values.nodes_info[i].node_rpc_port
+    MaxGasInvoke: 50
   Monitoring:
     Enabled: false
     Port: #@ data.values.nodes_info[i].node_monitoring_port


.docker/ir/go.protocol.template.yml
```diff
@@ -62,6 +62,7 @@ ApplicationConfiguration:
     Enabled: true
     EnableCORSWorkaround: false
     Port: #@ data.values.nodes_info[i].node_rpc_port
+    MaxGasInvoke: -1
```
roman-khimov, line 65: MaxGasInvoke: 50?

I mean unlimited is a bit dangerous.

.docker/build/Dockerfile.sharp.sources
```diff
@@ -22,7 +22,7 @@ RUN wget -O /tmp/neo-cli.zip https://github.com/neo-project/neo-node/archive/${C
 
 # Build neo-modules from source into /Plugins folder (only plugin .dll and plugin config are included, if you need other dependant .dlls, see the next step)
 ENV MODULESBRANCH="v3.0.0-preview3-00"
-ENV MODULES="LevelDBStore RocksDBStore RpcServer"
+ENV MODULES="LevelDBStore RocksDBStore RpcServer RpcNep5Tracker ApplicationLogs"
```
roman-khimov, line 25: NACK. neo-bench is about performance testing and these plugins seriously affect it. Just try it. Yeah, neo-go has all of that functionality without plugins, but that's just the way it is. You can make it an option, but installing them by default is wrong.

roman-khimov (changes requested):

roman-khimov (approved):

PR #41, opened 2020-10-05
*: post-preview3 updates

AnnaShaleva:
🍁 Fetch resent go modules
	🍃 Update Go and SharpSources Dockerfile revisions
	🍂 Increase network fee for UpdatePolicy transaction

C# Dockerfile is not updated as there are no new releases since preview3.


roman-khimov:
Can we remove dump.acc and single.acc from the repository?

roman-khimov (approved):

PR #43, opened 2020-10-14
Makefile: move dump.txs to build directory

AnnaShaleva:
Should be merged after #42.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -125,25 +125,25 @@ $(BUILD_DIR)/dump.txs: deps cmd/gen/main.go
 		&& go run ./gen -out ../$@
 
 # Generate both block dumps used for tests.
-dumps: ../$(BUILD_DIR)/single.acc ../$(BUILD_DIR)/dump.acc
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -115,14 +115,14 @@ pull:
 	@docker pull $(HUB)-sharp:$(TAG)
 
 # Generate `dump.txs` (run it before any benchmarks)
-gen: dump.txs
+gen: ../$(BUILD_DIR)/dump.txs
```
roman-khimov, line 118: Why `../` here, btw?

AnnaShaleva: It's relative to `./cmd/` dir, because we have to change directory to `./cmd/` before running any of our go scripts.

roman-khimov: But Makefile is at the top directory, for it it's just `$(BUILD_DIR)/...`. This `../` should be accounted for in other places, not in the makefile target.

roman-khimov (approved):

Issue #49, opened 2020-12-07
Update for NEP-17 and contract hash changes


roman-khimov:
Now that neo-project/neo-modules#412 is merged we can try testing both nodes at their current versions again with appropriate NEP-17 changes to neo-bench itself.

PR #50, opened 2020-12-08
Update neo-go version

fyrchik:
Related #49 .
Need to wait until preview4 to update C# nodes.

### DIFF
--- a/.docker/build/Dockerfile.golang
+++ b/.docker/build/Dockerfile.golang
@@ -6,7 +6,7 @@ WORKDIR /neo-go
 # Install deps:
 RUN apk add --no-cache git
 
-ARG REV="5f21178fe958df1afa8047c341257a2f3218a167"
+ARG REV="44b4c929922240a91a535c054da6f071f6c7ad64"
 ARG REPO="github.com/nspcc-dev/neo-go"
 
 # Clone and build repo:
--- a/.docker/build/Dockerfile.sharp.sources.from_local_dependencies
+++ b/.docker/build/Dockerfile.sharp.sources.from_local_dependencies
@@ -20,11 +20,11 @@ RUN set -x \
 #           ├── neo-modules
 #           └── neo-vm
 # Use ENV variables to specify branch, version or revision to download.
-ENV CLIBRANCH="cc7307c5eef5c7f77772c037f4b7619dae55a6bb"
-ENV MODULESBRANCH="e8a69c460be235cd9f10de3d1a2667b150252a52"
+ENV CLIBRANCH="f387c0138e650efca0f38ba3d20db4e72baadc92"
+ENV MODULESBRANCH="8403137813da5ea7f04c8b8d46b24539fa259d6f"
 ENV MODULES="LevelDBStore RocksDBStore RpcServer"
 ENV NEOVMBRANCH="a14b934f5b5c8779164e3b96b276984b3a57ccd4"
-ENV NEOBRANCH="54eff7e3662085a86a5e6cfefdc3e019215830f4"
+ENV NEOBRANCH="309981cece2f45386f87340860e2227d70a0dd33"
 RUN mkdir /neo-project && \
     wget -O /tmp/neo-cli.zip https://github.com/neo-project/neo-node/archive/${CLIBRANCH}.zip && \
     unzip -q -d /neo-project/ /tmp/neo-cli.zip && \
--- a/Makefile
+++ b/Makefile
@@ -13,9 +13,9 @@ DC_SHARP_RPC=.docker/rpc/docker-compose.sharp.yml
 
 DF_GO=.docker/build/Dockerfile.golang
 DF_BENCH=.docker/build/Dockerfile.bench
-DF_SHARP=.docker/build/Dockerfile.sharp
+#DF_SHARP=.docker/build/Dockerfile.sharp
 #DF_SHARP=.docker/build/Dockerfile.sharp.sources.from_binaries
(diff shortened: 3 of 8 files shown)


roman-khimov:
>  Need to wait until preview4 to update C# nodes.

I don't think so, they should work fine with latest NEP-17 PRs merged. The trick here is exactly to make it all work as of current state of both nodes (and they're relatively synchronized at the moment).

fyrchik:
@roman-khimov the problem is that C# docker file requires release tarball, and I haven't found if built for master.

roman-khimov:
See Dockerfile.sharp.sources.from_binaries (and Dockerfile.sharp.sources.from_local_dependencies, though that might be more involved, it's OK to have just the first one for now).

fyrchik:
There are two unmerged PRs https://github.com/neo-project/neo-node/pull/681 https://github.com/neo-project/neo/pull/2083 . Because of incompatible container fails to be built. I'll try using `git` instead of downloading zip files.

roman-khimov:
Just leave neo-vm at compatible commit, it hasn't been changed functionally for quite some time and we certainly don't need .NET 5.0 at the moment.

fyrchik:
Updated `neo` to commit right before introduction of management contract and `neo-vm` right before dotnet 5.0 update. Everything builds, though nodes accepted by C# node are empty. Nothing is logged, currently investigating.

fyrchik:
Both nodes build and run bench. But C# node accepts only 512 tx per block. Seems like policy contract invoke somehow failed.

.docker/ir/docker-compose.single.sharp.yml
```diff
@@ -6,7 +6,7 @@ services:
       - stats
     image: registry.nspcc.ru/neo-bench/neo-sharp:bench
     logging:
-      driver: "none"
+      driver: "journald"
```
roman-khimov, line 9: NACK.

Makefile
```diff
@@ -13,9 +13,9 @@ DC_SHARP_RPC=.docker/rpc/docker-compose.sharp.yml
 
 DF_GO=.docker/build/Dockerfile.golang
 DF_BENCH=.docker/build/Dockerfile.bench
-DF_SHARP=.docker/build/Dockerfile.sharp
+#DF_SHARP=.docker/build/Dockerfile.sharp
 #DF_SHARP=.docker/build/Dockerfile.sharp.sources.from_binaries
```
roman-khimov, line 17: I thought this one is easier to make work actually.

roman-khimov:
> But C# node accepts only 512 tx per block.

Well, why is that?

fyrchik:
The problem is in invalid policy contract hash (on a commit previous to management contract introduction).

fyrchik:
The problem was in contract hashes. Commit for `neo-go` with fix is from https://github.com/nspcc-dev/neo-go/pull/1616 . Another thing is increased cost of `transfer` (TBD fix in `neo-go`).

.docker/build/Dockerfile.sharp.sources.from_local_dependencies
```diff
@@ -20,11 +20,11 @@ RUN set -x \
...
 #           └── neo-vm
 # Use ENV variables to specify branch, version or revision to download.
-ENV CLIBRANCH="cc7307c5eef5c7f77772c037f4b7619dae55a6bb"
-ENV MODULESBRANCH="e8a69c460be235cd9f10de3d1a2667b150252a52"
+ENV CLIBRANCH="92363c5e66bd6028f73c53ab77492a11c783262e"
+ENV MODULESBRANCH="c995e65164ae10a303ec7515af5a0e93dc493be2"
 ENV MODULES="LevelDBStore RocksDBStore RpcServer"
 ENV NEOVMBRANCH="a14b934f5b5c8779164e3b96b276984b3a57ccd4"
-ENV NEOBRANCH="54eff7e3662085a86a5e6cfefdc3e019215830f4"
+ENV NEOBRANCH="2e6350136dd9cd36cc0804845b2eedd75a053a7f"
```
AnnaShaleva, line 27: If building from local dependencies works, then building from binaries should work too. Could you also update Dockerfile.sharp.sources.from_binaries?

AnnaShaleva (changes requested):

.docker/build/Dockerfile.sharp.sources.from_local_dependencies
fyrchik: Fixed.

fyrchik:
Updated `neo-go` and `neo` to master, `neo-vm` is still one commit behind `.NET` update.

.docker/build/Dockerfile.golang
```diff
@@ -6,7 +6,7 @@ WORKDIR /neo-go
 # Install deps:
 RUN apk add --no-cache git
 
-ARG REV="5f21178fe958df1afa8047c341257a2f3218a167"
+ARG REV="44b4c929922240a91a535c054da6f071f6c7ad64"
```
AnnaShaleva, line 9: Could you, please, update Go revision to ecdf95d5d6d8fc648ebbece74f8294e62201b765? It is more recent than 44b4c929922240a91a535c054da6f071f6c7ad64 and contains mempool bugfix which is important for bench.

AnnaShaleva (changes requested):

AnnaShaleva (approved):

roman-khimov (approved):
If all the logic changes needed are in place, we'd rather merge it as is to unblock #52. Changing neo-go revision is rather trivial.

PR #52, opened 2020-12-15
cmd: introduce transactions queue

### LINKED ISSUE #48: Fix congested mempool handling
neo-go mempool subsystem is very efficient and can accept a lot of transactions quickly on a sufficiently powerful machine:
```
2020/12/02 20:29:00 Sent 590186 transactions in 18.254299807s
2020/12/02 20:29:00 RPS: 32331.341
2020/12/02 20:29:00 RPC Errors: 409813 / 69.438%
```
Which unfortunately leads to mempool overflows (even for 500K mempool) as bench threads can push transactions even faster than this. Growing mempool further than that is not an option (500K is already more than enough, actually smaller mempool would be useful for testing as we can only put 64K transactions into one block). But these OOM errors returned from the RPC server are just counted as final errors now, no retries are being made. Which for four nodes setup leads to even more funny things:
```
2020/12/02 20:34:25 Sent 50000 transactions in 18.051894163s
2020/12/02 20:34:25 RPS: 2769.792
2020/12/02 20:34:25 RPC Errors: 949999 / 1899.998%
```
50K accepted, 950K rejected (1900% is not exactly 146%, but seems to be close enough in spirit), the benchmark will only process these 50K transactions.

We need to retry requests that failed because of OOM error. It can either be done with a bit map for sent/not sent transactions (1M transactions is 1M bits) or with a circular pipeline that will fan-out transactions to workers and fan-in failed ones.

AnnaShaleva:
Close #48.

Depends on #50.

### DIFF
--- a/cmd/internal/worker.go
+++ b/cmd/internal/worker.go
@@ -423,5 +423,5 @@ func (d *doer) Sender(ctx context.Context) {
 		log.Println("All transactions have been sent successfully")
 	}
 
-	log.Printf("RPC Errors: %d / %0.3f%%", errCount, (float64(errCount)/float64(count))*100)
+	log.Printf("RPC Errors: %d / %0.3f%%", errCount, (float64(errCount)/float64(count+errCount))*100)
 }
--- a/cmd/internal/worker.go
+++ b/cmd/internal/worker.go
@@ -177,7 +177,7 @@ func NewWorkers(opts ...WorkerOption) (Worker, error) {
 		return nil, errors.New("workers count could not be empty")
 	case p.dump == nil:
 		return nil, errors.New("dump could not be empty")
-	case len(p.dump.Transactions) < 1:
+	case p.dump.TransactionsQueue.Len() < 1:
 		return nil, errors.New("txs could not be empty")
 	case len(p.dump.Hashes) < 1:
 		return nil, errors.New("hashes could not be empty")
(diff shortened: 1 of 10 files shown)


AnnaShaleva:
All things are done, so ready for review.

roman-khimov:
@fyrchik, please review.

cmd/internal/worker.go
```diff
@@ -233,12 +233,25 @@ loop:
...
+				return
+			}
+			if err := d.cli.SendTX(ctx, tx.(string)); err != nil {
+				if errors.Is(err, ErrMempoolOOM) {
+					err := d.dump.TransactionsQueue.Put(tx.(string))
+					if err != nil {
+						log.Printf("failed to re-enqueue transaction: %s\n", err)
+						d.countErr.Inc()
+					}
+					time.Sleep(100 * time.Millisecond)
```
fyrchik, line 251: Why 100 milliseconds?

fyrchik (approved):

cmd/internal/worker.go
roman-khimov: This actually was discussed off-github, when the mempool is full sending transactions to it just puts unnecessary load on the system (it will only accept them after the new block), but of course we need to keep trying, just at lower pace until we can continue pushing at full speed. Given our 1-5s block time, 0.1s seems to be good enough choice, it doesn't put a lot of stress to already congested mempool yet at the same time it has an acceptable delay between block acceptance and resuming transaction sends.

PR #60, opened 2020-12-29
cmd: update dependencies for proper CPU usage counting

AnnaShaleva:
### Problem

Bench CPU usage sometimes can be more than 100%, e.g.:
```
MillisecondsFromStart, CPU, Mem
10027.374, 46.869%, 204.113MB
20055.605, 112.934%, 659.109MB
30089.750, 104.996%, 1039.277MB
40148.315, 76.385%, 1565.094MB
50209.257, 94.113%, 2313.957MB
60247.439, 105.199%, 2320.629MB
```


roman-khimov:
I really don't like these `576.770%` numbers, they need to be divided for the number of cores. 100% is all (be it 8 or 32) cores busy, simple as that.

AnnaShaleva:
@roman-khimov, the reason was in old dependencies, see ref in the commit description.

cmd/go.mod
```diff
@@ -5,19 +5,24 @@ go 1.13
 require (
 	github.com/Microsoft/go-winio v0.4.14 // indirect
 	github.com/Workiva/go-datastructures v1.0.50
+	github.com/containerd/containerd v1.4.3 // indirect
 	github.com/docker/distribution v2.7.1+incompatible // indirect
-	github.com/docker/docker v1.13.1
+	github.com/docker/docker v20.10.3+incompatible
```
roman-khimov, line 10: Some stable version available with the fix?

AnnaShaleva: There's a v19.03.15, but `go get github.com/moby/moby@v19.03.15` gives me
```
go: github.com/moby/moby v19.03.15 => v17.12.0-ce-rc1.0.20210128214336-420b1d36250f+incompatible
```
And I woud expect something like 
```
go: github.com/moby/moby v19.03.15 => v19.03.15...
```
Am I doing something wrong?

roman-khimov: It shouldn't be this way, but have you tried `v20.10.3`?

cmd/internal/docker.go
```diff
@@ -85,9 +86,9 @@ func NewStats(ctx context.Context, opts ...StatOption) (DockerStater, error) {
 		opts[i](p)
 	}
 
-	cli, err := client.NewEnvClient()
+	cli, err := client.NewClientWithOpts(client.WithVersion("1.40")) // version mey need to be downgraded on different hosts
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("see cmd/internal/docker.go for error details: %w", err)
```
roman-khimov, line 91: Doesn't seem to be user-friendy.

cmd/internal/docker.go
```diff
@@ -85,9 +86,9 @@ func NewStats(ctx context.Context, opts ...StatOption) (DockerStater, error) {
 		opts[i](p)
 	}
 
-	cli, err := client.NewEnvClient()
+	cli, err := client.NewClientWithOpts(client.WithVersion("1.40")) // version mey need to be downgraded on different hosts
```
roman-khimov, line 89: What this `1.40` refers to? How widespread is it? Are newer versions compatible with it?

cmd/go.mod
AnnaShaleva: `go: github.com/moby/moby v20.10.3` gives me `github.com/docker/docker v20.10.3+incompatible` in go.mod. And it works fine with this.

cmd/internal/docker.go
AnnaShaleva: It's a REST API version of the docker daemon engine / docker client. You can get it via `docker version`:
```
$ docker version
Client: Docker Engine - Community
 Version:           19.03.8
 API version:       1.40
 Go version:        go1.12.17
 Git commit:        afacb8b7f0
 Built:             Wed Mar 11 01:25:46 2020
 OS/Arch:           linux/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.8
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.12.17
  Git commit:       afacb8b7f0
  Built:            Wed Mar 11 01:24:19 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.2.13
  GitCommit:        7ad184331fa3e55e52b890ea95e65ba581ae3429
 runc:
  Version:          1.0.0-rc10
  GitCommit:        dc9208a3303feef5b3839f4323d9beb36df0a9dd
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```
You're right, it might not suite to someone if it has an old docker installed, but it's better to specify something. 

The most latest is 1.41.

cmd/internal/docker.go
AnnaShaleva: What should we write here?

cmd/go.mod
roman-khimov: OK, if the revision is correct and stable it should be good enough.

cmd/internal/docker.go
roman-khimov: If newer versions are compatible it's fine to have 1.40 there for now (I do have this API version also).

cmd/internal/docker.go
roman-khimov: `docker client init: `? I think it should write something meaningful in `%w` anyway on version mismatch.

roman-khimov (approved):

PR #80, opened 2021-08-17
Allow to benchmark GAS transfers

### LINKED ISSUE #77: GAS transfer bench
GAS is quite a different beast, it should work faster actually.

fyrchik:
Close #77 .
Current syntax: `make BENCH_TYPE=gas gen` (default is `BENCH_TYPE=neo`).


roman-khimov (commented):
We better have two separate tx dumps for two test modes. Then it'd be easy to switch between them (no regeneration needed).

Does it differ much from NEO on your machine?

roman-khimov:
Also I'm thinking about removing `make` from test runs, the more options like this we have the more it's inappropriate for the task.

roman-khimov:
> Does it differ much from NEO on your machine?

Just some 3% faster on mine. I'd expect to see bigger difference based on the fact that moving NEO actually also triggers GAS distribution so it's like moving two tokens with one transaction. And NEO balance decoding/encoding costs much more than GAS. Looks like memory management optimizations are still very much relevant for the node (but not critical now).

Makefile
```diff
@@ -117,7 +119,7 @@ $(BUILD_DIR)/dump.txs: cmd/gen/main.go
 	@echo "=> Generate transactions dump"
 	@set -x \
 		&& cd cmd/ \
-		&& go run ./gen -out ../$@
+		&& go run ./gen -type $(BENCH_TYPE) -out ../$@
```
AnnaShaleva, line 122: I think this feature and default behaviour should be mentioned in README.

fyrchik:
I have found a problem with this: for some reason pprof still shows NEO transfers. Investigating.

fyrchik:
This happened after rebase on  #87 because I forgot to provide `-i` arguments to `runner.sh`. Fixed this in last commit.

roman-khimov:
NEOBENCH_LOGGER should default to none to avoid this:
```
$ NEOBENCH_TYPE=NEP17 make start.GoSingle10wrk
...
WARNING: The NEOBENCH_LOGGER variable is not set. Defaulting to a blank string.
...
2021/08/29 12:18:59 Init 10 workers / 5m0s time limit (1000000 txs will try to send)
2021/08/29 12:19:00 empty block: 11
2021/08/29 12:19:00 empty block: 12
2021/08/29 12:19:01 CPU: 17.995%, Mem: 137.863MB
2021/08/29 12:19:03 CPU: 0.002%, Mem: 137.414MB
2021/08/29 12:19:05 CPU: 0.002%, Mem: 137.090MB
2021/08/29 12:19:07 CPU: 0.002%, Mem: 138.012MB
2021/08/29 12:19:09 CPU: 0.002%, Mem: 137.984MB
2021/08/29 12:19:11 CPU: 0.416%, Mem: 138.242MB
```
Also, when running as `NEOBENCH_TYPE=NEP17 make start.GoSingle10wrk` it regenerates tx file (same thing with GAS).

Makefile
```diff
@@ -93,15 +94,17 @@ pull:
 	@docker pull $(HUB)-go:$(TAG)
 	@docker pull $(HUB)-sharp:$(TAG)
 
-# Generate `dump.txs`
-gen: $(BUILD_DIR)/dump.txs
+gen:
+	NEOBENCH_TYPE=NEO make gen.NEO
+	NEOBENCH_TYPE=GAS make gen.GAS
+	NEOBENCH_TYPE=NEP17 make gen.NEP17
```
roman-khimov, line 100: Please don't call `make` from make. It's trivial to handle with dependencies.

fyrchik:
@roman-khimov have your tried the last commit? `NEOBENCH_LOGGER` should be taken from `.env` and the problem with regeneration should also be fixed now (at least it is ` Read 1000000 txs from /dump.txs` on my laptop). The regeneration right before the docker-compose is expected (an is current master or 3.0.2 behaviour).

Makefile
fyrchik: It was hard with environment variables, fixed with wildcards.

Makefile
```diff
@@ -93,15 +94,17 @@ pull:
...
 
-# IGNORE: create transactions dump
-$(BUILD_DIR)/dump.txs: cmd/gen/main.go
+# Generate `dump.txs`
+$(BUILD_DIR)/dump.%.txs: cmd/gen/main.go
 	@echo "=> Generate transactions dump"
 	@set -x \
 		&& cd cmd/ \
-		&& go run ./gen -out ../$@
+		&& go run ./gen -type $* -out ../$(BUILD_DIR)/dump.$*.txs
```
roman-khimov, line 107: `-out ../$@`?

Makefile
```diff
@@ -133,10 +136,10 @@ config:
 
 
 # Generate transactions, dump and nodes configurations for four-nodes network
-prepare: stop $(BUILD_DIR)/dump.txs $(BUILD_DIR)/dump.acc
+prepare: stop gen.$(NEOBENCH_TYPE) $(BUILD_DIR)/dump.acc
 
 # Generate transactions, dump and nodes configurations fore single-node network
-prepare.single: stop $(BUILD_DIR)/dump.txs $(BUILD_DIR)/single.acc
+prepare.single: stop gen.$(NEOBENCH_TYPE) $(BUILD_DIR)/single.acc
```
roman-khimov, line 142: It should depend on `$(BUILD_DIR)/dump.$(NEOBENCH_TYPE).txs`, because `gen.$(...)` should be PHONY.

cmd/bench/main.go
```diff
@@ -160,7 +160,7 @@ func main() {
 		dump = internal.ReadDump(in)
 		count = int(dump.TransactionsQueue.Len())
 	} else {
-		dump = internal.Generate(ctx, count)
+		dump = internal.Generate(ctx, internal.NEOTransfer, count)
```
roman-khimov, line 164: Throw error here? I'd expect dump creation be the standard prerequisite. Otherwise it should take `NEOBENCH_TYPE` into account.

cmd/dump/tokencontract/token.go
```diff
@@ -0,0 +1,72 @@
+package tokencontract
```
roman-khimov, line 1: Can we avoid copying this contract?

fyrchik: It is a bit different than example contract: we have a different owner and mint funds in `_deploy`. We certainly can avoid having it in repo, but I think it is not a big deal.

roman-khimov: If there are any updates we'll have to change two contracts.

Makefile
fyrchik: Fixed

cmd/bench/main.go
```diff
@@ -23,11 +23,6 @@ import (
 // - start sending txes to the node
 // - measure how much TX could be sent
 
-const (
-	coefficient = 1.3
-	defaultRate = 1300
-)
```
roman-khimov, line 29: What's that? Why in 9b8192e584968bfd2332e71d800cccb3b8d93c53?

cmd/bench/main.go
```diff
@@ -158,10 +148,8 @@ func main() {
 
 	if in := v.GetString("in"); in != "" {
 		dump = internal.ReadDump(in)
-		count = int(dump.TransactionsQueue.Len())
 	} else {
-		log.Println("Transactions dump was not provided, generating new transactions.")
-		dump = internal.Generate(ctx, internal.NEOTransfer, count)
+		log.Fatalf("Transactions dump file wasn't specified.")
```
roman-khimov, line 152: It's better to rework db11cad0638fe70c0c77632960774cd35fda1141 then, it has nothing to do with 9b8192e584968bfd2332e71d800cccb3b8d93c53 anyway.

Makefile
```diff
@@ -133,10 +133,10 @@ config:
 
 
 # Generate transactions, dump and nodes configurations for four-nodes network
-prepare: stop $(BUILD_DIR)/dump.txs $(BUILD_DIR)/dump.acc
+prepare: stop $(BUILD_DIR)/dump.$(NEOBENCH_TYPE).txs  $(BUILD_DIR)/dump.acc
 
 # Generate transactions, dump and nodes configurations fore single-node network
-prepare.single: stop $(BUILD_DIR)/dump.txs $(BUILD_DIR)/single.acc
+prepare.single: stop $(BUILD_DIR)/dump.$(NEOBENCH_TYPE).txs  $(BUILD_DIR)/single.acc
```
roman-khimov, line 139: Additional spaces are not really needed.

cmd/dump/main.go
```diff
@@ -154,6 +200,17 @@ func fillChain(bc *core.Blockchain, c *signer) error {
 		return err
 	}
 
+	// We deploy contract from priv to avoid having different hashes for single/4-node benchmarks.
+	txDeploy, err := newDeployTx(bc, priv, "./dump/tokencontract/token.nef",
```
roman-khimov, line 204: Of course you've written where this NEF comes from and how to regenerate it if needed?

roman-khimov (changes requested):

cmd/bench/main.go
fyrchik: `count` is no longer used, these numbers too (used for generation before).

roman-khimov (approved):

PR #82, opened 2021-08-18
build: use 3.0.1 modules for 3.0.2 C# node

roman-khimov:
I should've checked this, but image built from sources of course works
fine. It doesn't work from released binaries though as there is no 3.0.2
release of neo-modules. So fix it.


AnnaShaleva (approved):

AnnaShaleva:
> The image builds successfully BTW, it just doesn't work (RPC server is critical for us).

It's strange, because RPC server itself works and gives proper responses to `getblockcount` (I've checked manually by curl).

AnnaShaleva:
I've just found that it's a problem of `sharp.healthcheck.sh`.

PR #86, opened 2021-08-24
.docker: add network delay

fyrchik:
Delaying is done inside the container via `tc` command.
It can be enabled via environment $DELAY variable:
`DELAY=1000ms make start.GoFourNodes10wrk`

For this to work container needs CAP_NET_ADMIN capability.

Signed-off-by: Evgenii Stratonikov <evgeniy@nspcc.ru>

### DIFF
--- a/.docker/build/Dockerfile.golang
+++ b/.docker/build/Dockerfile.golang
@@ -29,6 +29,8 @@ RUN set -x \
 # Executable layer
 FROM alpine
 
+RUN apk add --no-cache iproute2-tc
+
 WORKDIR /neo-go
 
 WORKDIR /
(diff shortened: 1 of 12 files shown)


fyrchik:
Another possibility is to add delay directly to the network interface on the host, but this needs `tc` command on the host machine and will work also for the RPC node.

The behaviour is quite interesting: on my laptop TPS goes down for ~15%, but block times change from 5 seconds to ~20seconds (for 1 second delay to test if everything works).

fyrchik:
DELAY can be used for arbitrary parameters to tc: `DELAY="1000ms 100ms distribution normal" make start.GoFourNodes10wrk`

I think it is ok to use such variable inside container, but it is not very user-friendly. We can change this in #4 .

roman-khimov:
Please rebase it to include 1.17 changes.

roman-khimov (approved):

PR #87, opened 2021-08-25
Run benchmarks via script

### LINKED ISSUE #4: Drop make usage for test runs
It may be useful for builds (but that is not obvious for this particular case), but uncommenting something to run the test is just plain wrong, we need some trivial `./run.sh -with -some -flags` to drive test execution.

fyrchik:
1. Dockerfile fixes: Unify format and actually use declared network. 
2. Move all parameter hanling in `runner.sh`.

Close #4.


.docker/ir/docker-compose.sharp.yml
```diff
@@ -1,5 +1,13 @@
 version: '2.4'
 
+networks:
+  default:
+    name: neo_go_network
+    ipam:
+      config:
+        - subnet: 172.200.0.0/24
```
roman-khimov, line 8: It's gonna use the same subnet for every run?

fyrchik: Yes, but this is not a problem unless runs are performed in parallel (they can't currently).

roman-khimov: Actually, that's good, makes the system more predictable, knowing exactly where to knock at with pprof request is already a big improvement.

fyrchik:
So with last commit I have tried to make sensible defaults so that default invocation is just `./runner.sh -w 30`.
We can even try to allow providing empty list of arguments but this needs some refactoring.

.docker/ir/.env
```diff
@@ -0,0 +1 @@
+LOGGING=none
```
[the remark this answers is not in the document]
fyrchik: I think this is already documented in help, but we still have ability to use environment, so I also created `ENV.md`.
Fixed.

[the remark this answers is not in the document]
fyrchik: Moved to `README.md`.

roman-khimov (approved):

Issue #78, opened 2021-08-12
NEO voter bench


roman-khimov:
Voter GAS distribution is more complex, so seeing how it performs is also interesting.

roman-khimov:
Fixed by #90.

Issue #84, opened 2021-08-21
inmemory DB support


roman-khimov:
Currently we can't bench memory-based DB because of `db restore` that adds some blocks before the test, if it adds them to `inmemory` DB all changes get lost before the node actually starts. We can fix it by dropping dumps completely and instead pushing the changes needed using regular RPC client, it just takes a pair of blocks, so not a big problem to do while testing.

roman-khimov:
Related to #70, btw.

roman-khimov:
This also would allow the node to warm up before the test.

PR #90, opened 2021-09-01
Add some bench targets

### LINKED ISSUE #5: Single address vs. multiple addresses tests
After #3 we need to extend our test environment to use different types of transfers depending on some configuration parameter. There is a difference between transferring assets in 1:1, 1:N and N:N scenarios, so we need to cover that in neo-bench.

fyrchik:
Close #5, #78,

TODO: code cleanup


cmd/internal/gen.go
```diff
@@ -56,42 +63,56 @@ var SourceWifs = []string{
...
+		toAddr[i][0] = byte(i)
+		toAddr[i][util.Uint160Size-1] = byte(i)
+	}
+
 	w := io.NewBufBinWriter()
-	emit.AppCall(w.BinWriter,
-		contractHash, "transfer", callflag.All,
-		fromAddressHash, fromAddressHash, int64(1), nil)
-	emit.Opcodes(w.BinWriter, opcode.ASSERT)
+	for i := 0; i < n; i++ {
```
roman-khimov, line 140: That's certainly not the way it was supposed to work. Single transaction should still have a single transfer inside, it's just that instead of using one address in every transaction we use a number of them.

cmd/internal/gen.go
```diff
@@ -47,10 +47,19 @@ const (
 	ContractTransfer = "nep17"
 )
 
+// SourceWifs represents the senders of NEO and GAS.
+var SourceWifs = []string{
```
roman-khimov, line 60: How about 1000 or 10K addresses?

README.md
```diff
@@ -324,6 +324,7 @@ Name|Description|Default|Example
 NEOBENCH_LOGGER|Container logging facility|`none`|`none`, `journald`, `syslog`
 NEOBENCH_TC|Parameters passed to the `tc qdisc` (netem discipline) on container startup| |`delay 100ms`
 NEOBENCH_TYPE|Type of the load|`NEO`|`NEO`, `GAS`, `NEO-2-3` (transfer NEO from 2 senders to 3 recepients)
+NEOBENCH_VOTE|Vote for validators before the bench|empty|`1` or empty
```
roman-khimov, line 327: I'm not sure what's the difference. Registration only vs. registration with votes? Does it change anything? Can we always do it?

cmd/dump/main.go
```diff
@@ -234,6 +234,40 @@ func fillChain(bc *core.Blockchain, c *signer) error {
 	return addBlock(bc, c)
 }
 
+func newRegisterTx(bc *core.Blockchain, c *signer, p *keys.PrivateKey) *transaction.Transaction {
+	w := io.NewBufBinWriter()
+	emit.AppCall(w.BinWriter, bc.GoverningTokenHash(), "registerCandidate",
+		callflag.All, p.PublicKey().Bytes())
+	if w.Err != nil {
+		panic(w.Err)
+	}
```
AnnaShaleva, line 287: Maybe it's worth to add an `ASSERT`?

README.md
fyrchik: Registration is done for `VOTE` load to be applicable. `Vote` in dump is done for the `Transfer` loads.

cmd/internal/gen.go
fyrchik: Currently not supported.
The problem is that we transfer funds in chain dump, but send during tx generation. WIFs must be the same.
So we either should store all 1000 wifs in code/file or generate them on the fly in deterministic fashion.

README.md
roman-khimov: I mean you can always register some candidates, it doesn't hurt.

cmd/internal/gen.go
roman-khimov: But that's what we actually need, storing them in a file is not a big deal. And if we're to solve #84 it'd be easy to do initial transfers using dump's data.

fyrchik:
I have rebased this on master:
1. Added some info to the tx dump header, wifs are generated on the fly.
2. Remove bench for `vote` call performance.
3. More appropriate parameter handling (3 parameters instead of `NEO-10-22`).

TODO: test with sharp RPC.

fyrchik:
Checked C# RPC with voting and custom sender count, everything looks good.

cmd/gen/main.go
```diff
@@ -13,6 +13,9 @@ var (
 	out = flag.String("out", "./dump.txs", "Path to dump transactions.")
 	cnt = flag.Int("cnt", 1_000_000, "Count of txs that would be generated.")
 	typ = flag.String("type", internal.NEOTransfer, "Type of txs that would be generated.")
+
+	fromCount = flag.Int("from", 1, "Amount of tx senders")
+	toCount   = flag.Int("to", 1, "Amount of tx recepients")
```
roman-khimov, line 18: This actually changes 1-1 from using the same account in `from` and `to` to using two separate accounts.

cmd/internal/settings.go
```diff
@@ -94,6 +94,8 @@ func InitSettings() *viper.Viper {
 		"``Path to input file to load transactions.\n"+
 			"Example: -i ./dump.txs --in /path/to/import/transactions")
 
+	flags.BoolP("vote", "", false, "Vote befor the bench.")
```
roman-khimov, line 97: before

roman-khimov (changes requested):

cmd/gen/main.go
```diff
@@ -23,7 +24,12 @@ func main() {
 	case inp != nil && *inp != "":
 		internal.ReadDump(*inp)
 	case out != nil && *out != "" && cnt != nil && *cnt > 0:
-		internal.WriteDump(ctx, *out, *typ, *cnt)
+		p, _ := keys.NewPrivateKeyFromWIF("KxhEDBQyyEFymvfJD96q8stMbJMbZUb6D1PmXqBWZDU2WvbvVs9o")
```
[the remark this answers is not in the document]
fyrchik: Yes, this key is used in the deployed contract.

cmd/internal/gen.go
```diff
@@ -129,7 +139,8 @@ func Generate(ctx context.Context, opts BenchOptions, callback ...GenerateCallba
 				log.Fatal(ctx.Err())
 			}
 
-			txCh <- tx
+			index := rand.Intn(len(txR))
```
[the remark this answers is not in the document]
fyrchik: Made this more deterministic.
I don't it collecting them in order is important, as we have some randomness during tx send already (we send transactions in parallel).

cmd/gen/main.go
fyrchik: Fixed. Now first receivers are taken from senders and others are deterministically generated.

cmd/internal/gen.go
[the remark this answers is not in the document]
roman-khimov: Still, initial order can affect the result (as transactions are read and tried to be sent in order) and it'll probably be a bit different from run to run (and on different machines). Some bias could be added by these variations, so we better avoid it.

roman-khimov (commented):
Tx generation non-determinism still worries me.

cmd/internal/gen.go
```diff
@@ -96,30 +101,52 @@ func Generate(ctx context.Context, opts BenchOptions, callback ...GenerateCallba
...
-		tx = newTransferTx(opts.Senders[0], h)
-	default:
-		panic(fmt.Sprintf("invalid type: %s", opts.TransferType))
+	// We support both N-to-1 and 1-to-N cases, thus these index calculations.
+	max := len(opts.Senders)
+	if max < opts.ToCount {
+		max = opts.ToCount
+	}
+
+	txR := make([]txRequest, max)
```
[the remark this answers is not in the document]
fyrchik: Sending funds from multiple accounts? I don't think some symmetry is essential here (and if needed it can be achieved with different parameters, e.g. 6/3 or 6/12).

fyrchik:
Implemented deterministic generation in the last commit (checked with `sha256sum` for `GAS.1.1` dump).

roman-khimov (approved):

PR #93, opened 2021-09-08
bench: prepare nodes at runtime

fyrchik:
Sharp RPC node doesn't work for some reason (even with `ApplicationLogs` plugin installed.


runner.sh
```diff
@@ -169,7 +169,7 @@ if [ -z "$SINGLE" ]; then
       ;;
   esac
 
-  if [ "$RPC_TYPE" = go ]; then
+  if [ "$RPC_TYPE" = go ] || [ "$RPC_TYPE" = mixed ]; then
```
roman-khimov, line 172: But do we need to support "mixed" for RPC? We only have one RPC node currently. CNs are mixed, but RPC is not.

.docker/build/Dockerfile.sharp
```diff
@@ -50,7 +50,7 @@ RUN wget -O /opt/neo-cli.zip ${URL} && \
     rm -r /tmp/neo-cli && \
     rm /opt/neo-cli.zip
 
-ENV MODULES="DBFTPlugin RocksDBStore LevelDBStore RpcServer"
+ENV MODULES="DBFTPlugin RocksDBStore LevelDBStore RpcServer ApplicationLogs"
```
roman-khimov, line 53: Ouch. That would seriously affect C# testing. Can we check for tx side-effects like balances or contracts instead of execution result?

fyrchik: Well we can do an ad-hoc check:
1. For transfers check balance.
2. For deploy check contract state.

runner.sh
fyrchik: We set `RPC_TYPE` set to `IR_TYPE` by default. We can try set it to `go` by default (which is reasonable as it works better).

roman-khimov: `RPC_TYPE == IR_TYPE` is correct by default, but we can set it to `go` in case `IR_TYPE` is `mixed`.

cmd/go.mod
```diff
@@ -5,6 +5,7 @@ go 1.14
 require (
 	github.com/Workiva/go-datastructures v1.0.53
 	github.com/containerd/containerd v1.5.5 // indirect
+	github.com/davecgh/go-spew v1.1.1
```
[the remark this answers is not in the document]
fyrchik: Fixed.

roman-khimov (approved):

PR #100, opened 2021-09-22
plot: make figure a bit larger

roman-khimov:
Maybe it's 4K monitor, maybe it's trying to plot many lines at once, but I
feel current image size to be somewhat small.


fyrchik (approved):

AnnaShaleva (approved):
We can also use `plt.figure(dpi=1200)` before saving to increase resolution, something like that https://stackoverflow.com/questions/39870642/matplotlib-how-to-plot-a-high-resolution-graph.

PR #104, opened 2021-10-05
Use `version.Protocol` for auto-configuration

### LINKED ISSUE #94: Use Protocol.MillisecondsPerBlock for bench auto-configuration
After nspcc-dev/neo-go#2161 and neo-project/neo-modules#639 we can get this value from RPC node, so any code can use it instead of always assuming magic 5 and 1 values.

fyrchik:
Fix #94 .


roman-khimov (commented):
Update C# node built from sources and try this change there too?

AnnaShaleva (changes requested):
We have one more cludge I'd like to get rid of (and we can do it now). See the way how MsPerBlock is evaluated in the bench code: https://github.com/nspcc-dev/neo-bench/blob/95e17634eac7097143b67609928b7c1f263352e7/cmd/bench/main.go#L58.

fyrchik:
I have recently updated docker-compose to 2.0 and for some reason sharp nodes just hang. If I disable helthcheck, everything works fine on the latest commits.
@roman-khimov @AnnaShaleva could you check if sharp nodes setup is working for you (on this branch, see the last commit).

AnnaShaleva (commented):
Please, include the update of C# Modules revisions in this PR for both `Dockerfile.sharp.sources.from_binaries` and `Dockerfile.sharp.sources.from_local_dependencies`, otherwise these configurations are broken because of determined zero validators count.

I tried to run 4-nodes setup with `Dockerfile.sharp.sources.from_binaries` on Module's https://github.com/neo-project/neo-modules/commit/87ed9f0461271c4787705e5579a54c74f911b3c8 commit. It also hangs on my machine, but I don't think it is connected with your docker-compose update. This problem is known since https://github.com/nspcc-dev/neo-bench/pull/82#issuecomment-901656448.

So when I change the C# healthcheck to a dummy script the bench works fine and blocks are accepted. It's strange because released binaries worked fine with unchanged healthcheck script (I've checked this right after https://github.com/nspcc-dev/neo-bench/pull/82)

cmd/internal/prepare.go
```diff
@@ -167,19 +154,13 @@ func newNEP5Transfer(validatorCount int, sc util.Uint160, from, to util.Uint160,
 	return tx
 }
 
-func fillChain(ctx context.Context, c *client.Client, sgn *signer, vote bool, opts BenchOptions) error {
+func fillChain(ctx context.Context, c *client.Client, v *result.Version, sgn *signer, vote bool, opts BenchOptions) error {
```
roman-khimov, line 157: You can pass `result.Protocol` only, no need for full `result.Version`.

AnnaShaleva (approved):

cmd/internal/prepare.go
fyrchik: Fixed.

PR #105, opened 2021-10-05
runner.sh: help text update

AnnaShaleva:
Minor documentation update after https://github.com/nspcc-dev/neo-bench/pull/102.


roman-khimov:
But `golang:1.17-alpine` change is strange one to me. Looks like some interim issue, it shouldn't happen (the `tc` package should still be available) and we don't actually need Go components in runtime.

PR #108, opened 2021-10-16
Fix C# healthcheck script

### LINKED ISSUE #107: Take care of C# healthcheck
After C# node update in #104, it stopped working properly due to healthcheck issues.

fyrchik:
FIx #107 .


AnnaShaleva (approved):

.docker/build/sharp.healthcheck.sh
```diff
@@ -1,37 +1,38 @@
 #!/bin/bash
 
 maxDelayBlocks=10
-export port=`jq -r '.PluginConfiguration.Servers[0].Port' < /neo-cli/Plugins/RpcServer/config.json`
+port=$(jq -r '.PluginConfiguration.Servers[0].Port' < /neo-cli/Plugins/RpcServer/config.json)
+export port
```
roman-khimov, line 5: I guess I'll start hating shellcheck soon. This change makes zero sense to me.

.docker/build/sharp.healthcheck.sh
```diff
@@ -1,37 +1,38 @@
...
-    block=`curl -s -X POST http://$node -H 'Content-Type: application/json' -d '{ "jsonrpc": "2.0", "id": 5, "method": "getblockcount", "params": [] }'| jq '.result'`
+    block=$(curl -s -X POST "http://$node" -H 'Content-Type: application/json' -d '{ "jsonrpc": "2.0", "id": 5, "method": "getblockcount", "params": [] }'| jq '.result')
 
     if [ "$block" == "" ]
     then
         block=0
     fi
 
-    syncDelay=`expr $block - $curBlock`
+    syncDelay=$(( block - curBlock ))
```
roman-khimov: And that is a clear bashism, isn't it?

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,16 @@
...
+    steps:
+      - uses: actions/checkout@v1
+      - name: Run the sh-checker
+        uses: luizm/action-sh-checker@master
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SHELLCHECK_OPTS: -x # allow to source .env
+          SHFMT_OPTS: -s # arguments to shfmt.
+        with:
+          sh_checker_comment: true
```
roman-khimov, line 16: No newline.

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,16 @@
+name: shellcheck
+on:
+  - pull_request
+jobs:
+  sh-checker:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Run the sh-checker
```
roman-khimov, line 9: Do we need it? Some things are nice there, but some...

.docker/build/sharp.healthcheck.sh
fyrchik: It is, though we already have bash after shebang.

.github/workflows/shellcheck.yml
fyrchik: I agree it is questionable. However, we can disable some annoying hints both in comments and in `.yml` file.
The benefit is that at least we are of possible quirks. We don't change this scripts often and setting up this in IDE is easy, though.

fyrchik: Ok, dropped this.

roman-khimov (approved):

PR #121, opened 2022-06-29
README: update usage example

AnnaShaleva:
And adjust `make test` target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,12 +60,7 @@ build.node.sharp:
 	@docker build -q -t $(HUB)-sharp:$(TAG) -f $(DF_SHARP) $(BUILD_DIR)
 
 # Test local benchmark (go run) with Neo single node
-test: single.go
-	@echo "=> Test Single node"
-	@set -x \
-		&& cd cmd/ \
-		&& go run ./bench -o ../single.log -i ../$(BUILD_DIR)/dump.$(NEOBENCH_TYPE).txs -d "SingleNode" -m rate -q 1000 -z 1m -t 30s -a localhost:20331
-	@make stop
+test: start.GoSingle30wrk
 
 # Bootup NeoGo single node
 single.go: stop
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -62,10 +62,7 @@ build.node.sharp:
 # Test local benchmark (go run) with Neo single node
 test: single.go
 	@echo "=> Test Single node"
-	@set -x \
-		&& cd cmd/ \
-		&& go run ./bench -o ../single.log -i ../$(BUILD_DIR)/dump.$(NEOBENCH_TYPE).txs -d "SingleNode" -m rate -q 1000 -z 1m -t 30s -a localhost:20331
-	@make stop
+	./runner.sh --validators 1 -d "GoSingle" -m wrk -w 100 -z 5m -t 30s
```
roman-khimov, line 65: `test` can just be an alias of `start.GoSingle100wrk` then and I'd use 30 worker threads since that's the setup used in Medium posts.

AnnaShaleva: Fixed.

Makefile
```diff
@@ -62,10 +62,7 @@ build.node.sharp:
 # Test local benchmark (go run) with Neo single node
 test: single.go
 	@echo "=> Test Single node"
-	@set -x \
-		&& cd cmd/ \
-		&& go run ./bench -o ../single.log -i ../$(BUILD_DIR)/dump.$(NEOBENCH_TYPE).txs -d "SingleNode" -m rate -q 1000 -z 1m -t 30s -a localhost:20331
-	@make stop
+	make start.GoSingle30wrk
```
roman-khimov, line 65: It should be `$(MAKE) start...`, but we're likely to have other misuses of `make`...

Makefile
```diff
@@ -62,10 +62,7 @@ build.node.sharp:
 # Test local benchmark (go run) with Neo single node
 test: single.go
```
roman-khimov, line 63: ... so we can just do `test: start.GoSingle30wrk` here, because it already has all of the dependencies needed and we don't need yet another line in the output.

AnnaShaleva: You're right, fixed.

PR #142, opened 2023-11-03
.github: add building and testing workflows

### LINKED ISSUE #141: Build/test workflow
## Is your feature request related to a problem? Please describe.
I'm always frustrated when I don't see that code in a PR can't even be built, let alone run, like in #140.

## Describe the solution you'd like
At least a workflow to build things, but some minimal runtime sanity checks would be useful as well.

AnnaShaleva:
- Ensure that every PR doesn't break at least the building make target. Manual dispatch is also supported.
- Add linter, CodeQL and testing jobs.
- Add codeowners.
- Fix linting issues.

Close #141.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,86 @@
+name: Tests
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches:
+      - master
+    types: [opened, synchronize]
+    paths-ignore:
+      - '**/*.md'
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: 'Ref to test Bench, GoNode and SharpNode images [default: latest master; examples: 0a4ff9d3e4a9ab432fd5812eb18c98e03b5a7432]'
+        required: false
+        default: ''
+
+jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/setup-go@v4
+        with:
+          go-version: '1.19'
+          cache-dependency-path: cmd/go.sum
+
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.inputs.ref }}
+
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v3
+        with:
+          version: latest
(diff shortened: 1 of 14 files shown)


AnnaShaleva:
Let's wait until the building job is finished before the merge.

AnnaShaleva:
Wow, we already have https://github.com/nspcc-dev/neo-bench/issues/141 for it. Let's then add a testing job also.

cmd/internal/client.go
```diff
@@ -127,7 +128,8 @@ func (c *RPCClient) SendTX(ctx context.Context, tx string) error {
 	rpc := fmt.Sprintf(`{"jsonrpc": "2.0", "id": 1, "method": "sendrawtransaction", "params": ["%s"]}`, tx)
 
 	if err := c.doRPCCall(ctx, rpc, &res, c.txSender); err != nil {
-		if respErr, ok := err.(*neorpc.Error); ok && (respErr.Message == "The memory pool is full and no more transactions can be sent." || respErr.Message == "OutOfMemory") {
+		msg := err.Error()
+		if strings.Contains(msg, "The memory pool is full and no more transactions can be sent.") || strings.Contains(msg, "OutOfMemory") {
```
AnnaShaleva, line 132: Ref. https://github.com/nspcc-dev/neo-go/issues/3188.

AnnaShaleva: And we can't use the direct cast here, because it will fail in case of wrapped error.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,86 @@
...
+
+jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/setup-go@v4
+        with:
+          go-version: '1.19'
```
roman-khimov, line 27: Why not 1.21?

roman-khimov (approved):

.github/workflows/tests.yml
AnnaShaleva: The same reason as for https://github.com/nspcc-dev/neo-go/pull/2997.

PR #164, opened 2024-04-17
config: add `--msPerBlock` flag

AliceInHunterland:
`./runner.sh --nodes mixed --rpc sharp -d "Mytest" -m rate -q 50 -z 5m -t 30s --time $(TIME_PER_BLOCK)`

example:
```
TARGETS := start.GoFourNodes10wrk start.GoFourNodes30wrk start.GoFourNodes100wrk
TIME_BLOCKS := 5000 1000 500 200 100

run-block-tests:
	@$(foreach target,$(TARGETS), \
		$(foreach time,$(TIME_BLOCKS), \
			echo "Running $(target) with time per block $(time) ms" && \
			$(MAKE) $(target) TIME_PER_BLOCK=$(time); \
		) \
	)
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ BUILD_DIR=.docker/build
 NEOBENCH_TYPE ?= NEO
 NEOBENCH_FROM_COUNT ?= 1
 NEOBENCH_TO_COUNT ?= 1
+MS_PER_BLOCK ?= 0
 
 .PHONY: help
 
@@ -112,7 +113,7 @@ config:
 	@echo "=> Generate configurations for single-node and four-nodes networks from templates"
 	@set -x \
 		&& cd ./cmd \
-		&& go run ./config/ --go-template go.protocol.template.yml --go-db leveldb --sharp-template sharp.protocol.template.yml --sharp-db LevelDBStore
+		&& go run ./config/ --go-template go.protocol.template.yml --go-db leveldb --sharp-template sharp.protocol.template.yml --sharp-db LevelDBStore --msPerBlock $(MS_PER_BLOCK)
 
 
 # Generate transactions, dump and nodes configurations for four-nodes network
--- a/README.md
+++ b/README.md
@@ -296,6 +296,9 @@ The following default configurations are available:
                                     Example: -l journald -l syslog -l json-file
        --tc                         Arguments to pass to 'tc qdisc netem' inside the container.
                                     Example: 'delay 100ms'
+       --msPerBlock                 Protocol setting specifying the minimal (and targeted for) time interval between blocks. Must be an integer number of milliseconds.
+                                    The default value is set in configuration templates and is 1s and 5s for single node and multinode setup respectively.
+                                    Example: --msPerBlock 3000
 
 ```
 
--- a/cmd/config/main.go
+++ b/cmd/config/main.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/k14s/ytt/pkg/cmd/template"
 	"github.com/nspcc-dev/neo-go/pkg/config"
@@ -29,6 +30,7 @@ var (
 	goDB              = flag.String("go-db", "leveldb", "database for Go node")
 	sharpTemplateFile = flag.String("sharp-template", "", "configuration template file for C# node")
 	sharpDB           = flag.String("sharp-db", "LevelDBStore", "database for C# node")
+	msPerBlock        = flag.Int("msPerBlock", 0, "time per block in milliseconds")
 )
 
 func main() {
@@ -115,6 +117,9 @@ func generateGoConfig(templatePath, database, suffix string) error {
 			return fmt.Errorf("unable to decode node template #%d: %w", i, err)
 		}
 		template.ApplicationConfiguration.DBConfiguration.Type = database
+		if msPerBlock != nil && *msPerBlock > 0 {
+			template.ProtocolConfiguration.TimePerBlock = time.Duration(*msPerBlock) * time.Millisecond
+		}
 		var configFile string
 		nodeName, err := nodeNameFromSeedList(template.ApplicationConfiguration.P2P.Addresses, template.ProtocolConfiguration.SeedList)
 		if err != nil {
@@ -154,6 +159,9 @@ func generateSharpConfig(templatePath, storageEngine, suffix string) error {
 			return fmt.Errorf("unable to decode node template #%d: %w", i, err)
 		}
 		template.ApplicationConfiguration.Storage.Engine = storageEngine
+		if msPerBlock != nil && *msPerBlock > 0 {
+			template.ProtocolConfiguration.MillisecondsPerBlock = *msPerBlock
+		}
 		var configFile string
 		nodeName, err := nodeNameFromSeedList([]string{":" + strconv.Itoa(int(template.ApplicationConfiguration.P2P.Port))}, template.ProtocolConfiguration.SeedList)
 		if err != nil {
--- a/runner.sh
+++ b/runner.sh
@@ -51,6 +51,9 @@ show_help() {
 	echo "                                    Example: -l journald -l syslog -l json-file"
 	echo "       --tc                         Arguments to pass to 'tc qdisc netem' inside the container."
 	echo "                                    Example: 'delay 100ms'"
+	echo "       --msPerBlock                 Protocol setting specifying the minimal (and targeted for) time interval between blocks. Must be an integer number of milliseconds."
+	echo "                                    The default value is set in configuration templates and is 1s and 5s for single node and multinode setup respectively."
+	echo "                                    Example: --msPerBlock 1000"
 	exit 0
 }
 
@@ -185,6 +188,12 @@ while test $# -gt 0; do
 	--tc)
 		test $# -gt 0 || fatal "tc arguments should be specified"
 		export NEOBENCH_TC="$1"
+		shift
+		;;
+
+	--msPerBlock)
+		test $# -gt 0 || fatal "milliseconds per block should be specified"
+		export MS_PER_BLOCK="$1"
 		shift
 		;;
 


runner.sh
```diff
@@ -50,6 +51,7 @@ show_help() {
 	echo "   -l, --log                        Enable logging on consensus nodes."
 	echo "       --tc                         Arguments to pass to 'tc qdisc netem' inside the container."
 	echo "                                    Example: 'delay 100ms'"
+	echo "   --time                           Set the time per block (in milliseconds)."
```
AnnaShaleva, line 54: It should be `--msPerBlock`. Adjust the commit message as far.

cmd/config/main.go
```diff
@@ -115,6 +117,7 @@ func generateGoConfig(templatePath, database, suffix string) error {
 			return fmt.Errorf("unable to decode node template #%d: %w", i, err)
 		}
 		template.ApplicationConfiguration.DBConfiguration.Type = database
+		template.ProtocolConfiguration.TimePerBlock = time.Duration(*timePerBlock) * time.Millisecond
```
AnnaShaleva, line 120: Needs to be changed depending on the decision in the comment above.

Makefile
```diff
@@ -112,7 +113,7 @@ config:
 	@echo "=> Generate configurations for single-node and four-nodes networks from templates"
 	@set -x \
 		&& cd ./cmd \
-		&& go run ./config/ --go-template go.protocol.template.yml --go-db leveldb --sharp-template sharp.protocol.template.yml --sharp-db LevelDBStore
+		&& go run ./config/ --go-template go.protocol.template.yml --go-db leveldb --sharp-template sharp.protocol.template.yml --sharp-db LevelDBStore --time-per-block $(MS_PER_BLOCK)
```
AnnaShaleva, line 116: These two names should be unified. Use `--msPerBlock` and `MS_PER_BLOCK`.

AnnaShaleva: It's literally the protocol configuration field name.

cmd/config/main.go
```diff
@@ -29,6 +30,7 @@ var (
 	goDB              = flag.String("go-db", "leveldb", "database for Go node")
 	sharpTemplateFile = flag.String("sharp-template", "", "configuration template file for C# node")
 	sharpDB           = flag.String("sharp-db", "LevelDBStore", "database for C# node")
+	timePerBlock      = flag.Int("time-per-block", -1, "time per block in milliseconds")
```
AnnaShaleva, line 33: Let's use `0` as default.

cmd/config/main.go
```diff
@@ -115,6 +117,9 @@ func generateGoConfig(templatePath, database, suffix string) error {
 			return fmt.Errorf("unable to decode node template #%d: %w", i, err)
 		}
 		template.ApplicationConfiguration.DBConfiguration.Type = database
+		if timePerBlock != nil && *timePerBlock > 0 {
+			template.ProtocolConfiguration.TimePerBlock = time.Duration(*timePerBlock) * time.Millisecond
+		}
```
AnnaShaleva, line 122: OK, maybe `0` default doesn't suite us. Have you checked, what happens if no flag is provided? What is the `timePerBlock` variable value then?

runner.sh
```diff
@@ -50,6 +55,8 @@ show_help() {
 	echo "   -l, --log                        Enable logging on consensus nodes."
 	echo "       --tc                         Arguments to pass to 'tc qdisc netem' inside the container."
 	echo "                                    Example: 'delay 100ms'"
+	echo "   --msPerBlock                     Protocol setting specifying the minimal (and targeted for) time interval between blocks. Must be an integer number of milliseconds."
```
AnnaShaleva, line 58: Fix the indent, this flag doesn't have the short name.

runner.sh
```diff
@@ -172,6 +179,12 @@ while test $# -gt 0; do
 		shift
 		;;
 
+  --msPerBlock)
+    test $# -gt 0 || fatal "time per block should be specified."
+    export TIME_PER_BLOCK="$1"
+    shift
+    ;;
+
```
AnnaShaleva, line 187: Fix the indent, please.

runner.sh
```diff
@@ -16,6 +16,11 @@ export NEOBENCH_FROM_COUNT=${NEOBENCH_FROM_COUNT:-1}
 export NEOBENCH_TC=${NEOBENCH_TC:-}
 export NEOBENCH_TO_COUNT=${NEOBENCH_TO_COUNT:-1}
 export NEOBENCH_VALIDATOR_COUNT=${NEOBENCH_VALIDATOR_COUNT:-4}
+if [ "$NEOBENCH_VALIDATOR_COUNT" -eq 1 ]; then
+    export TIME_PER_BLOCK=${TIME_PER_BLOCK:-1s}
```
AnnaShaleva, line 20: Make it `MS_PER_BLOCK`.

AnnaShaleva (commented):
The main question is: there are two steps during every benchmark run: the first is node configuration files generation from YAML templates, and the second is nodes start based on the generated configuration. MsPerBlock setting is added to both these steps. Have you ensured that these two settings are consistent? What happens if I'll specify different values at these two steps?

cmd/config/main.go
AliceInHunterland: if no flag is provided the default `MS_PER_BLOCK` will be used
```
if [ "$NEOBENCH_VALIDATOR_COUNT" -eq 1 ]; then
    export MS_PER_BLOCK=${MS_PER_BLOCK:-1s}
else
    export MS_PER_BLOCK=${MS_PER_BLOCK:-5s}
fi
```

AliceInHunterland:
> The main question is: there are two steps during every benchmark run: the first is node configuration files generation from YAML templates, and the second is nodes start based on the generated configuration. MsPerBlock setting is added to both these steps. Have you ensured that these two settings are consistent? What happens if I'll specify different values at these two steps?

```
Before make prepare 1000
=> Stop environment
=> Generate configurations for single-node and four-nodes networks from templates
+ cd ./cmd
+ go run ./config/ --go-template go.protocol.template.yml --go-db leveldb --sharp-template sharp.protocol.template.yml --sharp-db LevelDBStore --msPerBlock 1000
creating: 1963633258/go.protocol.template.yml
creating: 1963633258/go.protocol.template.yml
creating: 1963633258/sharp.protocol.template.yml
creating: 1963633258/sharp.protocol.template.yml
After make prepare 1000
```
if you specify  different value after `make prepare` directly in config files it will use your new value. so for running test it will use values from the last step.

Makefile
```diff
@@ -10,6 +10,7 @@ BUILD_DIR=.docker/build
 NEOBENCH_TYPE ?= NEO
 NEOBENCH_FROM_COUNT ?= 1
 NEOBENCH_TO_COUNT ?= 1
+MS_PER_BLOCK ?= 1000
```
AnnaShaleva, line 13: > if you specify different value after make prepare directly in config files it will use your new value. so for running test it will use values from the last step.

This behaviour is OK. However, I don't like this line follows the https://github.com/nspcc-dev/neo-bench/pull/164#discussion_r1575158570, because it overrides the configuration template values. I think that there was a correct value in older version of this PR.

`MS_PER_BLOCK ?= 1000` sets default value of `MS_PER_BLOCK` to 1000 if this variable is not specified (and most of the time it's not specified). Then you pass this value to the `config` target:
```
&& go run ./config/ --go-template go.protocol.template.yml --go-db leveldb --sharp-template sharp.protocol.template.yml --sharp-db LevelDBStore --msPerBlock $(MS_PER_BLOCK)
```

Which means that template-based value will always be overriden by 1000ms irrespective of the setup:
https://github.com/nspcc-dev/neo-bench/blob/7c1946a3f50e718d0481b547201ba8976d12ddeb/.docker/ir/go.protocol.template.yml#L6-L16

And what I'd like to see here is no default (i.e. use template-based values).

runner.sh
```diff
@@ -16,6 +16,11 @@ export NEOBENCH_FROM_COUNT=${NEOBENCH_FROM_COUNT:-1}
 export NEOBENCH_TC=${NEOBENCH_TC:-}
 export NEOBENCH_TO_COUNT=${NEOBENCH_TO_COUNT:-1}
 export NEOBENCH_VALIDATOR_COUNT=${NEOBENCH_VALIDATOR_COUNT:-4}
+if [ "$NEOBENCH_VALIDATOR_COUNT" -eq 1 ]; then
+    export MS_PER_BLOCK=${MS_PER_BLOCK:-1s}
+else
+    export MS_PER_BLOCK=${MS_PER_BLOCK:-5s}
+fi
```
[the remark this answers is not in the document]
AliceInHunterland: Okay, you are right, adjusted it. Now I'm expecting this behaviour:
if  `--msPerBlock` is passed to `runner.sh` then  `MS_PER_BLOCK` is exported to `Makefile` and passed to `./config` -> configuration template value overwrites.
Otherwise `MS_PER_BLOCK` is 0 -> no overwriting.

README.md
```diff
@@ -296,7 +296,9 @@ The following default configurations are available:
                                     Example: -l journald -l syslog -l json-file
        --tc                         Arguments to pass to 'tc qdisc netem' inside the container.
                                     Example: 'delay 100ms'
-
+       --msPerBlock                 Protocol setting specifying the minimal (and targeted for) time interval between blocks. Must be an integer number of milliseconds.
+                                    Default value is set in configuration templates and are 1s for single node and 5s for 4 nodes setup.
+                                    Example: --msPerBlock 3000
```
AnnaShaleva, line 301: Missing new line in the end.

README.md
```diff
@@ -296,7 +296,9 @@ The following default configurations are available:
                                     Example: -l journald -l syslog -l json-file
        --tc                         Arguments to pass to 'tc qdisc netem' inside the container.
                                     Example: 'delay 100ms'
-
+       --msPerBlock                 Protocol setting specifying the minimal (and targeted for) time interval between blocks. Must be an integer number of milliseconds.
+                                    Default value is set in configuration templates and are 1s for single node and 5s for 4 nodes setup.
```
AnnaShaleva, line 300: > for 4 nodes

For 4 and 7 nodes, just replace it with "multinode setup".

runner.sh
```diff
@@ -188,6 +191,12 @@ while test $# -gt 0; do
 		shift
 		;;
 
+	--msPerBlock)
+		test $# -gt 0 || fatal "time per block should be specified"
```
AnnaShaleva, line 195: > time per block 

milliseconds per block

README.md
AnnaShaleva: > and are 1s

and is 1s

AnnaShaleva (commented):
The logic itself LGTM, minor style fixes are left.

README.md
AnnaShaleva: Not fixed.

README.md
```diff
@@ -296,7 +296,10 @@ The following default configurations are available:
...
```
AnnaShaleva, line 300: Fix the commit message: "`time-per-block` is used in generation config files."

AnnaShaleva (approved):

PR #167, opened 2024-04-22
*: migrate to Docker Compose V2

### LINKED ISSUE #166: Migrate to Docker Compose V2
Docker Compose V1 is deprecated since June 2023, we need to migrate all our scripts to Docker Compose V2 so that users with fresh Docker were able to run neo-bench painlessly. Ref. https://docs.docker.com/compose/migrate/.

AnnaShaleva:
Close #166.

### DIFF
--- a/.docker/ir/docker-compose.go.7.yml
+++ b/.docker/ir/docker-compose.go.7.yml
@@ -1,5 +1,3 @@
-version: '2.4'
-
 networks:
   default:
     name: neo_go_network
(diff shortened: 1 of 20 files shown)


AnnaShaleva:
OK, we're not going to add macOS-based bench job since Docker is not installed on the macOS GitHub runners by default, and manual installation causes a lot of problems.

roman-khimov (approved):

PR #174, opened 2024-05-16
cmd: make mempool OOM error handler more adaptive

AnnaShaleva:
Once mempool OOm error happens, worker should wait for a while until block processing finishes. Make this awaiting interval depend on MSPerBlock protocol setting so that it's possible to perform precise calculations for low MSPerBlock setting.

### DIFF
--- a/cmd/bench/main.go
+++ b/cmd/bench/main.go
@@ -33,15 +33,16 @@ func main() {
 	defer cancel()
 
 	var (
-		workers    int
-		rate       int
-		msPerBlock int
-		threshold  time.Duration
-		dump       *internal.Dump
-		desc       = v.GetString("desc")
-		timeLimit  = v.GetDuration("timeLimit")
-		mode       = internal.BenchMode(v.GetString("mode"))
-		client     *internal.RPCClient
+		workers         int
+		rate            int
+		msPerBlock      int
+		mempoolOOMDelay time.Duration
+		threshold       time.Duration
+		dump            *internal.Dump
+		desc            = v.GetString("desc")
+		timeLimit       = v.GetDuration("timeLimit")
+		mode            = internal.BenchMode(v.GetString("mode"))
+		client          *internal.RPCClient
 	)
 
 	switch mode {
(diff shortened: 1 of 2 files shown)


cmd/internal/worker.go
```diff
@@ -213,6 +221,8 @@ func (d *doer) worker(ctx context.Context, idx *atomic.Int64, start time.Time) {
 	var (
 		done  = ctx.Done()
 		timer = time.NewTimer(d.timeLimit)
+		// mempoolOOMDelay is the time interval to pause sender's work after mempool OOM error occurred on tx submission.
+		mempoolOOMDelay = time.Duration(d.msPerBlock / 50 * int64(time.Millisecond))
```
AnnaShaleva, line 225: @roman-khimov, the resulting value is the same for 5-seconds block (100 ms), but for 1-seconds blocks (and less) this value is lower than previously considered 100ms. To me, it  could be `time.Duration(d.msPerBlock / 10 * int64(time.Millisecond))` as far, not sure which one is better.

roman-khimov: There is no perfect value for it. It was heuristically chosen to not poll the node too often with useless requests and start pushing transactions ASAP once mempool is flushed. But your combination doesn't work for `msPerBlock < 50`.

cmd/internal/worker.go
```diff
@@ -75,6 +76,13 @@ func WorkerBlockchainClient(cli *RPCClient) WorkerOption {
 	}
 }
 
+// WorkerMSPerBlock sets milliseconds per block protocol setting.
+func WorkerMSPerBlock(msPerBlock int64) WorkerOption {
+	return func(p *doerParams) {
+		p.msPerBlock = msPerBlock
```
roman-khimov, line 82: You can push mempoolOOMDelay directly, calculate is once for all workers.

cmd/internal/worker.go
AnnaShaleva: How about the updated version?

cmd/bench/main.go
```diff
@@ -61,6 +62,11 @@ func main() {
 		log.Fatalf("could not receive RPC Node version: %v", err)
 	}
 	msPerBlock = version.Protocol.MillisecondsPerBlock
+	if msPerBlock > 1000 {
+		mempoolOOMDelay = time.Duration(int64(msPerBlock) / 50 * int64(time.Millisecond))
+	} else {
+		mempoolOOMDelay = time.Duration(int64(msPerBlock) / 10 * int64(time.Millisecond))
```
roman-khimov, line 68: `mempoolOOMDelay = time.Duration(msPerBlock) * time.Millisecond / 50`?

AnnaShaleva: Fixed, but I'd still suggest to preserve this 50/10 division.

Issue #175, opened 2024-05-21
Take care of C# node dockerfiles


AnnaShaleva:
https://github.com/neo-project/neo-modules/pull/912 is finally merged which means that https://github.com/neo-project/neo is a monorepo. We need to migrate from `Dockerfile.sharp.sources.from_binaries` and `Dockerfile.sharp.sources.from_local_dependencies` to a single Dockerfile that allows to build node from the specific revision from sources.

We also need to update `Dockerfile.sharp` to be able to build released monorepo version.

This issue is relevant for post-3.7.4 versions of Neo node. Depends on https://github.com/neo-project/neo/pull/3232 and https://github.com/neo-project/neo/pull/3244.

roman-khimov:
We need to digest 3.7.4 first, it already changes a lot.

AnnaShaleva:
> We need to digest 3.7.4 first, 

I'm in progress, this issue is for future development.

PR #172, opened 2024-05-14
bench: add workers into rate mode

AliceInHunterland:
Close #170

PR #169, opened 2024-05-03
*: add external flag for testing external networks

AliceInHunterland:
`./runner.sh -e -d "Go4x1" -m rate -q 1000 -z 5m -t 30s -a 192.168.1.100:20331 -a 192.168.1.101:20331`

Close #160

PR #176, opened 2024-05-21
Update bench to 3.7.4 and pre-0.106.0 version

AnnaShaleva:
@roman-khimov, I suggest not to touch `Dockerfile.sharp.sources.from_binaries` and `Dockerfile.sharp.sources.from_local_dependencies` because we need to remove one of them and completely refactor another one in the #175. We have `Dockerfile.sharp` that allows us to properly build the last non-mono released version, I think it's enough.

What do you think?


AnnaShaleva:
Also, we may wait until 0.106.0 is released, and I'll update this PR afterwards.

roman-khimov:
> I suggest not to touch Dockerfile.sharp.sources.from_binaries and Dockerfile.sharp.sources.from_local_dependencies

I thought about making an interim version, but probably you're right, it won't allow to build `master` anyway, so we better have stable pre-built versions for now, tag them and then rework the build subsequently. What we need is some way to test `master`, so #175 is very much relevant for us in the near future.

roman-khimov (commented):
Let's wait for 0.106.0 proper.

AnnaShaleva:
>Let's wait for 0.106.0 proper.

Updated, depends on #172 and #169.

roman-khimov (approved):

AnnaShaleva:
3.7.4 is tagged successfully, can move on with #175.

PR #181, opened 2024-08-20
Go 1.21

### LINKED ISSUE #135: Go 1.21 upgrade check list
Once we drop support for 1.20, take a look at nspcc-dev/neo-go#3089.

roman-khimov:
Fix #135.


AnnaShaleva (dismissed):

cmd/go.mod
```diff
@@ -1,19 +1,18 @@
...
 require (
-	github.com/Workiva/go-datastructures v1.1.1
-	github.com/docker/docker v26.1.4+incompatible
-	github.com/fatih/color v1.16.0
+	github.com/Workiva/go-datastructures v1.1.5
+	github.com/docker/docker v26.1.5+incompatible
+	github.com/fatih/color v1.17.0
 	github.com/k14s/ytt v0.30.0
-	github.com/mailru/easyjson v0.7.7
 	github.com/moby/moby v25.0.3+incompatible
```
AnnaShaleva, line 10: BTW, somehow it happened that `docker` dependency has greater version than `moby`. Likely https://github.com/nspcc-dev/neo-bench/pull/178 was merged without corresponding moby upgrade. We need to keep them aligned, let's update moby in the same PR if v26.1+ is available.

AnnaShaleva: https://github.com/moby/moby/releases/tag/v26.1.5

Yep, it's there for us.

roman-khimov: OK, but there is more fun with github.com/k14s/ytt (it's very aggressive and wants 1.22.something already) and there is also a problem with Docker 27+ where some APIs we use for statistics have changed.

roman-khimov: Fixed.

AnnaShaleva: Let's just include `github.com/k14s/ytt` upgrade to #153 and keep Docker-dependent things at 0.26+ for now, and eventually we'll diges these API changes.

AnnaShaleva (approved):

cmd/go.mod
roman-khimov: Yeah, I'll try to do this, but my previous attempt ended up with not just `go 1.22.smth` in `go.mod`, but also `toolchain 1.23.smth` there which really surprised me. But let's see how it goes with newer Go already present in go.mod.

AnnaShaleva: BTW, I also faced with this problem when trying to update crypto-related NeoGo dependencies.

PR #183, opened 2024-08-21
Go 1.22

roman-khimov:


cmd/internal/prepare.go
```diff
@@ -382,7 +382,7 @@ func awaitTx(ctx context.Context, duration time.Duration, check ...func() (bool,
 
 	attempts := duration / retryInterval
 
-	for i := 0; i < len(check); {
+	for i := range check {
```
AnnaShaleva, line 385: We can't do this, it changes the behaviour. If a call to `check[i]` failed, then we need to retry the same call one more time.

roman-khimov: I remember it was suspicious. OK, will fix.

roman-khimov: Reverted. Likely it can be improved, but not this time.

AnnaShaleva (approved):

PR #184, opened 2024-09-09
workflows: switch to org-wide golangci-lint job

roman-khimov:


.golangci.yml
```diff
@@ -1,73 +0,0 @@
-# This file contains all available configuration options
-# with their default values.
```
roman-khimov, line 2: One thing that bugs me is that this can affect local linter runs.

roman-khimov: A make target to download it from .github can be added.

AnnaShaleva (approved):

.golangci.yml
AnnaShaleva: It should be updated manually if any changes are included into .github's config. Other than that it works OK on my machine.

roman-khimov: Yeah, the only other way to sync is submodules, but it's _so_ inconvenient (especially the need to bump them).

roman-khimov: I think the main thing is that we're still guarded by GH linter job, so if local config is out of sync we'll still see linter job failure and it should be sufficient.

AnnaShaleva: Let’s keep the current way, it’s good.

tatiana-nspcc (approved):

PR #224, opened 2026-08-25
bench: add scenario with conflicting mempools

Turalchik:
Adds a benchmark scenario that sends pairs of conflicting transactions to random RPC nodes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,8 @@ help:
 	start.GoFourNodes10wrk start.GoFourNodes30wrk start.GoFourNodes100wrk \
 	start.GoFourNodes25rate start.GoFourNodes50rate start.GoFourNodes60rate start.GoFourNodes300rate start.GoFourNodes1000rate \
 	start.GoFourNodes4RPC10wrk start.GoFourNodes4RPC30wrk start.GoFourNodes4RPC100wrk \
+	start.GoFourNodes4RPCConflicts10wrk start.GoFourNodes4RPCConflicts30wrk start.GoFourNodes4RPCConflicts100wrk \
+	start.GoFourNodes4RPCConflicts25rate start.GoFourNodes4RPCConflicts50rate start.GoFourNodes4RPCConflicts60rate start.GoFourNodes4RPCConflicts300rate start.GoFourNodes4RPCConflicts1000rate \
 	start.SharpSingle10wrk start.SharpSingle30wrk start.SharpSingle100wrk \
 	start.SharpSingle25rate start.SharpSingle50rate start.SharpSingle60rate start.SharpSingle300rate start.SharpSingle1000rate \
 	start.SharpFourNodes10wrk start.SharpFourNodes30wrk start.SharpFourNodes100wrk \
@@ -120,6 +122,7 @@ pull:
 gen: $(BUILD_DIR)/dump.NEO.$(NEOBENCH_FROM_COUNT).$(NEOBENCH_TO_COUNT).txs
 gen: $(BUILD_DIR)/dump.GAS.$(NEOBENCH_FROM_COUNT).$(NEOBENCH_TO_COUNT).txs
 gen: $(BUILD_DIR)/dump.NEP17.$(NEOBENCH_FROM_COUNT).$(NEOBENCH_TO_COUNT).txs
+gen: $(BUILD_DIR)/dump.conflict.$(NEOBENCH_FROM_COUNT).$(NEOBENCH_TO_COUNT).txs
 
 # Generate `dump.txs`
 $(BUILD_DIR)/dump.%.$(NEOBENCH_FROM_COUNT).$(NEOBENCH_TO_COUNT).txs: cmd/gen/main.go
@@ -216,6 +219,32 @@ start.GoFourNodes4RPC30wrk:
 
 start.GoFourNodes4RPC100wrk:
 	./runner.sh --rpc go4 -d "Go4x4" -m wrk -w 100 -z 5m -t 30s
+
+#	## Conflicting transactions:
+start.GoFourNodes4RPCConflicts10wrk:
+	./runner.sh --rpc go4 -b conflict -d "Go4x4Conflict" -w 10 -z 5m -t 30s
+
+start.GoFourNodes4RPCConflicts30wrk:
+	./runner.sh --rpc go4 -b conflict -d "Go4x4Conflict" -w 30 -z 5m -t 30s
+
+start.GoFourNodes4RPCConflicts100wrk:
+	./runner.sh --rpc go4 -b conflict -d "Go4x4Conflict" -w 100 -z 5m -t 30s
+
+#	## Rate:
+start.GoFourNodes4RPCConflicts25rate:
+	./runner.sh --rpc go4 -b conflict -d "Go4x4Conflict" -m rate -q 25 -z 5m -t 30s
+
+start.GoFourNodes4RPCConflicts50rate:
+	./runner.sh --rpc go4 -b conflict -d "Go4x4Conflict" -m rate -q 50 -z 5m -t 30s
+
+start.GoFourNodes4RPCConflicts60rate:
+	./runner.sh --rpc go4 -b conflict -d "Go4x4Conflict" -m rate -q 60 -z 5m -t 30s
+
+start.GoFourNodes4RPCConflicts300rate:
+	./runner.sh --rpc go4 -b conflict -d "Go4x4Conflict" -m rate -q 300 -z 5m -t 30s
+
+start.GoFourNodes4RPCConflicts1000rate:
+	./runner.sh --rpc go4 -b conflict -d "Go4x4Conflict" -m rate -q 1000 -z 5m -t 30s
 
 ## Go×4 + SharpRPC
 #
--- a/cmd/internal/client.go
+++ b/cmd/internal/client.go
@@ -13,6 +13,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/nspcc-dev/neo-go/pkg/core"
 	"github.com/nspcc-dev/neo-go/pkg/core/block"
 	"github.com/nspcc-dev/neo-go/pkg/io"
 	"github.com/nspcc-dev/neo-go/pkg/neorpc"
@@ -44,6 +45,9 @@ const DefaultTimeout = time.Second * 30
 var (
 	// ErrMempoolOOM is returned from `sendrawtransaction` when node cannot process transaction due to mempool OOM.
 	ErrMempoolOOM = errors.New("node cannot process transaction due to mempool OOM")
+	// ErrConflict is returned from `sendrawtransaction` when the node rejects a
+	// transaction because it conflicts with mempool or chain content via the Conflicts attribute.
+	ErrConflict = errors.New("node rejected transaction due to a Conflicts attribute clash")
 )
 
 // NewRPCClient creates new client for RPC communications.
@@ -118,15 +122,24 @@ func (c *RPCClient) GetVersion(ctx context.Context) (*result.Version, error) {
 
 // SendTX sends transaction.
 func (c *RPCClient) SendTX(ctx context.Context, tx string) error {
+	idx := int(c.inc.Add(1) % c.len)
+	return c.SendTXToNode(ctx, tx, idx)
+}
+
+// SendTXToNode sends transaction to the node at nodeIdx.
+func (c *RPCClient) SendTXToNode(ctx context.Context, tx string, nodeIdx int) error {
 	var res struct {
 		Hash util.Uint256 `json:"hash"`
 	}
 	rpc := fmt.Sprintf(`{"jsonrpc": "2.0", "id": 1, "method": "sendrawtransaction", "params": ["%s"]}`, tx)
 
-	if err := c.doRPCCall(ctx, rpc, &res, c.txSender); err != nil {
+	if err := c.doRPCCallToAddr(ctx, rpc, &res, c.txSender, nodeIdx); err != nil {
 		msg := err.Error()
 		if errors.Is(err, neorpc.ErrMempoolCapReached) || strings.Contains(msg, "OutOfMemory") {
 			return ErrMempoolOOM
+		}
+		if strings.Contains(msg, core.ErrHasConflicts.Error()) {
+			return ErrConflict
 		}
 		return err
 	} else if res.Hash.Equals(util.Uint256{}) {
@@ -134,6 +147,14 @@ func (c *RPCClient) SendTX(ctx context.Context, tx string) error {
 	}
 
 	return nil
+}
+
+// PickNodePair returns two distinct node indices.
+func (c *RPCClient) PickNodePair() (int, int) {
+	base := c.inc.Add(2)
+	a := int(base % c.len)
+	b := int((base + 1) % c.len)
+	return a, b
 }
 
 // GetBlock sends getblock RPC request.
@@ -163,9 +184,14 @@ func (c *RPCClient) GetBlockCount(ctx context.Context) (int, error) {
 	return num, c.doRPCCall(ctx, rpc, &num, c.blockRequester)
 }
 
-func (c *RPCClient) doRPCCall(_ context.Context, call string, result any, client *fasthttp.Client) error {
-	idx := c.inc.Add(1) % c.len
+func (c *RPCClient) doRPCCall(ctx context.Context, call string, result any, client *fasthttp.Client) error {
+	idx := int(c.inc.Add(1) % c.len)
+	return c.doRPCCallToAddr(ctx, call, result, client, idx)
+}
 
+// doRPCCallToAddr sends call to c.addr[idx]. The caller must ensure
+// 0 <= idx < len(c.addr).
+func (c *RPCClient) doRPCCallToAddr(_ context.Context, call string, result any, client *fasthttp.Client, idx int) error {
 	req, res := fasthttp.AcquireRequest(), fasthttp.AcquireResponse()
 	defer func() {
 		fasthttp.ReleaseRequest(req)
--- a/cmd/internal/gen.go
+++ b/cmd/internal/gen.go
@@ -42,6 +42,8 @@ type (
 	txRequest struct {
 		tx  *transaction.Transaction
 		acc *wallet.Account
+		// conflict is tx's optional Conflicts-attribute counterpart.
+		conflict *transaction.Transaction
 	}
 )
 
@@ -52,6 +54,8 @@ const (
 	GASTransfer = "gas"
 	// ContractTransfer is the type of deployed NEP17 contract transfer tx.
 	ContractTransfer = "nep17"
+	// ConflictTransfer is the type of conflicting transaction pair.
+	ConflictTransfer = "conflict"
 )
 
 // newNEOTransferTx returns NEO transfer transaction with random nonce.
@@ -77,14 +81,17 @@ func newTransferTx(p *keys.PrivateKey, contractHash, toAddr util.Uint160) *trans
 		panic(w.Err)
 	}
 
-	script := w.Bytes()
+	return newTx(fromAddressHash, w.Bytes())
+}
+
+func newTx(account util.Uint160, script []byte) *transaction.Transaction {
 	tx := transaction.New(script, 15000000)
 	tx.NetworkFee = 1500000 // hardcoded for now
 	tx.ValidUntilBlock = 1200
-	tx.Signers = append(tx.Signers, transaction.Signer{
-		Account: fromAddressHash,
+	tx.Signers = []transaction.Signer{{
+		Account: account,
 		Scopes:  transaction.CalledByEntry,
-	})
+	}}
 	return tx
 }
 
@@ -95,11 +102,20 @@ func Generate(ctx context.Context, opts BenchOptions, callback ...GenerateCallba
 	start := time.Now()
 	count := int(opts.TxCount)
 
-	dump := Dump{
-		TransactionsQueue: queue.NewRingBuffer(opts.TxCount),
+	resultSize := 1
+	if strings.ToLower(opts.TransferType) == ConflictTransfer {
+		resultSize = 2
 	}
 
-	log.Printf("Generate %d txs", count)
+	dump := Dump{
+		TransactionsQueue: queue.NewRingBuffer(opts.TxCount * uint64(resultSize)),
+	}
+
+	if resultSize == 1 {
+		log.Printf("Generate %d txs", count)
+	} else {
+		log.Printf("Generate %d conflicting transaction pairs (%d txs)", count, count*resultSize)
+	}
 
 	txCh := make([]chan txRequest, genWorkerCount)
 	for i := range txCh {
@@ -107,13 +123,13 @@ func Generate(ctx context.Context, opts BenchOptions, callback ...GenerateCallba
 	}
 	result := make([]chan txBlob, genWorkerCount)
 	for i := range result {
-		result[i] = make(chan txBlob, 1)
+		result[i] = make(chan txBlob, resultSize)
 	}
 
 	var wg sync.WaitGroup
 	for i := range genWorkerCount {
 		wg.Go(func() {
-			genTxWorker(i, txCh[i], result[i])
+			genTxWorker(txCh[i], result[i])
 		})
 	}
 
@@ -132,7 +148,10 @@ func Generate(ctx context.Context, opts BenchOptions, callback ...GenerateCallba
 			}
 		}
 
-		var tx *transaction.Transaction
+		var (
+			tx       *transaction.Transaction
+			conflict *transaction.Transaction // an optional conflicting tx.
+		)
 		switch strings.ToLower(opts.TransferType) {
 		case NEOTransfer:
 			tx = newNEOTransferTx(sender, receiver)
@@ -141,10 +160,15 @@ func Generate(ctx context.Context, opts BenchOptions, callback ...GenerateCallba
 		case ContractTransfer:
 			h, _ := util.Uint160DecodeStringLE("ceb508fc02abc2dc27228e21976699047bbbcce0")
 			tx = newTransferTx(sender, h, receiver)
+		case ConflictTransfer:
+			senderHash := sender.GetScriptHash()
+			tx = newTx(senderHash, []byte{byte(opcode.RET)})
+			conflict = newTx(senderHash, []byte{byte(opcode.RET)})
 		default:
 			panic(fmt.Sprintf("invalid type: %s", opts.TransferType))
 		}
 		txR[i].tx = tx
+		txR[i].conflict = conflict
 		txR[i].acc = wallet.NewAccountFromPrivateKey(sender)
 	}
 
@@ -155,7 +179,26 @@ func Generate(ctx context.Context, opts BenchOptions, callback ...GenerateCallba
 				log.Fatal(ctx.Err())
 			}
 
-			txCh[i%len(txCh)] <- txR[i%len(txR)]
+			wrkIdx := uint32(i % len(txCh))
+			baseNonce := wrkIdx << 24 // 255 possible workers and 16M transactions should be enough
+			r := txR[i%len(txR)]
+
+			tx := *r.tx
+			tx.Nonce = baseNonce | uint32(2*i)
+
+			task := txRequest{tx: &tx, acc: r.acc}
+
+			if r.conflict != nil {
+				conflict := *r.conflict
+				conflict.Nonce = baseNonce | uint32(2*i+1)
+				conflict.Attributes = []transaction.Attribute{{
+					Type:  transaction.ConflictsT,
+					Value: &transaction.Conflicts{Hash: tx.Hash()},
+				}}
+				task.conflict = &conflict
+			}
+
+			txCh[wrkIdx] <- task
 		}
 		for _, ch := range txCh {
 			close(ch)
@@ -164,16 +207,18 @@ func Generate(ctx context.Context, opts BenchOptions, callback ...GenerateCallba
 	}()
 
 	for i := range count {
-		r := <-result[i%len(result)]
+		for range resultSize {
+			r := <-result[i%len(result)]
 
-		err := dump.TransactionsQueue.Put(r.blob)
-		if err != nil {
-			log.Fatalf("Cannot enqueue transaction #%d: %s", i, err)
-		}
+			err := dump.TransactionsQueue.Put(r.blob)
+			if err != nil {
+				log.Fatalf("Cannot enqueue transaction #%d: %s", i, err)
+			}
 
-		for j := range callback {
-			if err := callback[j](r.hash, r.blob); err != nil {
-				log.Fatalf("Callback returns error: %d %v", i, err)
+			for j := range callback {
+				if err := callback[j](r.hash, r.blob); err != nil {
+					log.Fatalf("Callback returns error: %d %v", i, err)
+				}
 			}
 		}
 	}
@@ -188,31 +233,30 @@ func Generate(ctx context.Context, opts BenchOptions, callback ...GenerateCallba
 	return &dump
 }
 
-func genTxWorker(n int, ch <-chan txRequest, out chan<- txBlob) {
-	baseNonce := n << 24 // 255 possible workers and 16M transactions should be enough
-	i := 0
-
+func genTxWorker(ch <-chan txRequest, out chan<- txBlob) {
 	buf := io.NewBufBinWriter()
 	for tr := range ch {
-		tx := *tr.tx
-		tx.Nonce = uint32(baseNonce | i)
+		out <- signAndEncode(buf, tr.acc, tr.tx)
 
-		if err := tr.acc.SignTx(netmode.PrivNet, &tx); err != nil {
-			log.Fatalf("Could not sign tx: %v", err)
+		if tr.conflict != nil {
+			out <- signAndEncode(buf, tr.acc, tr.conflict)
 		}
+	}
+}
 
-		buf.Reset()
-		tx.EncodeBinary(buf.BinWriter)
+func signAndEncode(buf *io.BufBinWriter, acc *wallet.Account, tx *transaction.Transaction) txBlob {
+	if err := acc.SignTx(netmode.PrivNet, tx); err != nil {
+		log.Fatalf("Could not sign tx: %v", err)
+	}
 
-		if buf.Err != nil {
-			log.Fatalf("Could not prepare transaction: %d %v", i, buf.Err)
-		}
+	buf.Reset()
+	tx.EncodeBinary(buf.BinWriter)
+	if buf.Err != nil {
+		log.Fatalf("Could not prepare transaction: %v", buf.Err)
+	}
 
-		out <- txBlob{
-			hash: tx.Hash().String(),
-			blob: base64.StdEncoding.EncodeToString(buf.Bytes()),
-		}
-
-		i++
+	return txBlob{
+		hash: tx.Hash().String(),
+		blob: base64.StdEncoding.EncodeToString(buf.Bytes()),
 	}
 }
--- a/cmd/internal/reader.go
+++ b/cmd/internal/reader.go
@@ -4,6 +4,7 @@ import (
 	"compress/gzip"
 	"log"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/Workiva/go-datastructures/queue"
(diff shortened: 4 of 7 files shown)


cmd/bench/main.go
```diff
@@ -174,7 +174,26 @@ func main() {
...
+		log.Printf("Running conflicting-transactions scenario: %d pairs", pairs)
+
+		opts := internal.ConflictOptions{
+			Senders:         dump.BenchOptions.Senders,
+			PairsCount:      pairs,
+			NetworkFee:      v.GetInt64("conflict-network-fee"),
+			ValidUntilBlock: blk.Index + 1000,
+		}
+
+		generated := internal.GenerateConflictPairs(opts)
```
AnnaShaleva: Why here? Generate first via `make gen` like it's done for other bench scenarious. Here you only need to read ready-to-use dump.

cmd/internal/client.go
```diff
@@ -118,12 +118,25 @@ func (c *RPCClient) GetVersion(ctx context.Context) (*result.Version, error) {
...
+}
+
+func (c *RPCClient) SendTXToNode(ctx context.Context, tx string, nodeIdx int) error {
+	return c.sendTXToAddr(ctx, tx, nodeIdx)
+}
+
+func (c *RPCClient) AddrCount() int {
+	return len(c.addr)
+}
+
```
AnnaShaleva: Not needed, revert.

Makefile
```diff
@@ -217,6 +218,10 @@ start.GoFourNodes4RPC30wrk:
 start.GoFourNodes4RPC100wrk:
 	./runner.sh --rpc go4 -d "Go4x4" -m wrk -w 100 -z 5m -t 30s
 
+#	## Conflicting transaction pairs
+start.GoFourNodes4RPC1000000PairConflictTransactions:
```
AnnaShaleva: Need readable name.

cmd/internal/worker.go
```diff
@@ -18,6 +18,7 @@ type (
 		Wait()
 		Prepare(ctx context.Context, vote bool, opts BenchOptions)
 		Sender(ctx context.Context)
+		SendConflictPairs(ctx context.Context, pairs []ConflictPair)
```
AnnaShaleva: Diverges from existing methods.

cmd/bench/main.go
```diff
@@ -174,7 +174,26 @@ func main() {
 	log.Printf("Started test from block = %v at unix time = %v", blk.Index, blk.Timestamp)
 
 	go wrk.Parser(ctx, blk)
-	go wrk.Sender(ctx)
+
+	if pairs := v.GetInt("conflict-pairs"); pairs > 0 {
```
AnnaShaleva: `make gen` has the `type` argument, it should be used to generate proper type of items.

cmd/internal/conflict.go
```diff
@@ -0,0 +1,204 @@
...
+const (
+	conflictNetworkFee      int64  = 2_000_000
+	conflictValidUntilBlock uint32 = 1200
+)
+
+func generateConflictPairs(ctx context.Context, opts BenchOptions, callback ...GenerateCallback) *Dump {
+	if len(opts.Senders) == 0 {
+		log.Fatal("conflict scenario: no senders available")
+	}
+	if opts.TxCount%2 != 0 {
```
AnnaShaleva: TxCount is the number of pairs in this case, don't need to devide by 2.

cmd/internal/conflict.go
```diff
@@ -0,0 +1,204 @@
...
+		log.Fatalf("conflict scenario: tx count must be even to form pairs, got %d", opts.TxCount)
+	}
+
+	pairsCount := int(opts.TxCount) / 2
+	dump := Dump{TransactionsQueue: queue.NewRingBuffer(opts.TxCount)}
+
+	log.Printf("Generating %d conflicting transaction pairs (%d transactions)", pairsCount, opts.TxCount)
+
+	for i := range pairsCount {
```
AnnaShaleva: Should be done exactly as done by the standard tx generator: run parallel routines, generate a single pair per a single task. It should be possible to squash Generate/generateConflictPairs logic.

cmd/internal/conflict.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+	b, err := d.dump.TransactionsQueue.Get()
+	if err != nil {
+		return "", "", false
+	}
+
+	return a.(string), b.(string), true
+}
+
+func (d *doer) sendConflictTx(ctx context.Context, blob string, nodeIdx int, start time.Time) {
```
AnnaShaleva: Should also be possible to unify this code with an existing worker's code. Let me check and try it out.

AnnaShaleva:
Rebase is needed.

cmd/internal/client.go
```diff
@@ -118,16 +121,32 @@ func (c *RPCClient) GetVersion(ctx context.Context) (*result.Version, error) {
...
 	}
 	rpc := fmt.Sprintf(`{"jsonrpc": "2.0", "id": 1, "method": "sendrawtransaction", "params": ["%s"]}`, tx)
 
-	if err := c.doRPCCall(ctx, rpc, &res, c.txSender); err != nil {
+	if err := c.doRPCCallToAddr(ctx, rpc, &res, c.txSender, idx); err != nil {
 		msg := err.Error()
 		if errors.Is(err, neorpc.ErrMempoolCapReached) || strings.Contains(msg, "OutOfMemory") {
 			return ErrMempoolOOM
 		}
+		if strings.Contains(msg, "has conflicts") {
```
AnnaShaleva: Can we use errors.Is? We have designated `neorpc` error for this.

cmd/internal/gen.go
```diff
@@ -90,33 +94,88 @@ func newTransferTx(p *keys.PrivateKey, contractHash, toAddr util.Uint160) *trans
 
 var genWorkerCount = runtime.NumCPU()
 
-// Generate used to generate the specified number of transactions.
+// Generate runs opts.TxCount tasks in parallel, one task per worker
+// iteration: a single transfer transaction per task for the standard
+// transfer types, or a conflicting pair per task for ConflictTransfer.
 func Generate(ctx context.Context, opts BenchOptions, callback ...GenerateCallback) *Dump {
```
AnnaShaleva, line 101: It's unnecessary complication. Don't change `Generate` too significantly, it's not necessary to create two separate `txSetBuilder`s. I prefer to introduce as less changes to the existing codebase as possible. We need to refactor generator part, and here's the idea (`git diff` comparing to the current master):
```
anna@kiwi:~/Documents/GitProjects/nspcc-dev/neo-bench$ git diff
diff --git a/cmd/internal/gen.go b/cmd/internal/gen.go
index 1d5f73e..87b494c 100644
--- a/cmd/internal/gen.go
+++ b/cmd/internal/gen.go
@@ -40,8 +40,9 @@ type (
        }
 
        txRequest struct {
-               tx  *transaction.Transaction
-               acc *wallet.Account
+               tx       *transaction.Transaction
+               conflict *transaction.Transaction
+               acc      *wallet.Account
        }
 )
 
@@ -52,6 +53,7 @@ const (
        GASTransfer = "gas"
        // ContractTransfer is the type of deployed NEP17 contract transfer tx.
        ContractTransfer = "nep17"
+       ConflictTransfer = "conflict"
 )
 
 // newNEOTransferTx returns NEO transfer transaction with random nonce.
@@ -106,8 +108,12 @@ func Generate(ctx context.Context, opts BenchOptions, callback ...GenerateCallba
                txCh[i] = make(chan txRequest, 1)
        }
        result := make([]chan txBlob, genWorkerCount)
+       resultSize := 1
+       if opts.TransferType == ConflictTransfer {
+               resultSize = 2
+       }
        for i := range result {
-               result[i] = make(chan txBlob, 1)
+               result[i] = make(chan txBlob, resultSize)
        }
 
        var wg sync.WaitGroup
@@ -132,7 +138,10 @@ func Generate(ctx context.Context, opts BenchOptions, callback ...GenerateCallba
                        }
                }
 
-               var tx *transaction.Transaction
+               var (
+                       tx       *transaction.Transaction
+                       conflict *transaction.Transaction // an optional conflicting tx.
+               )
                switch strings.ToLower(opts.TransferType) {
                case NEOTransfer:
                        tx = newNEOTransferTx(sender, receiver)
@@ -141,10 +150,13 @@ func Generate(ctx context.Context, opts BenchOptions, callback ...GenerateCallba
                case ContractTransfer:
                        h, _ := util.Uint160DecodeStringLE("ceb508fc02abc2dc27228e21976699047bbbcce0")
                        tx = newTransferTx(sender, h, receiver)
+               case ConflictTransfer:
+                       tx, conflict = newConflictTxPair(sender, receiver)
                default:
                        panic(fmt.Sprintf("invalid type: %s", opts.TransferType))
                }
                txR[i].tx = tx
+               txR[i].conflict = conflict
                txR[i].acc = wallet.NewAccountFromPrivateKey(sender)
        }
 
@@ -164,16 +176,18 @@ func Generate(ctx context.Context, opts BenchOptions, callback ...GenerateCallba
        }()
 
        for i := range count {
-               r := <-result[i%len(result)]
+               for range resultSize {
+                       r := <-result[i%len(result)]
 
-               err := dump.TransactionsQueue.Put(r.blob)
-               if err != nil {
-                       log.Fatalf("Cannot enqueue transaction #%d: %s", i, err)
-               }
+                       err := dump.TransactionsQueue.Put(r.blob)
+                       if err != nil {
+                               log.Fatalf("Cannot enqueue transaction #%d: %s", i, err)
+                       }
 
-               for j := range callback {
-                       if err := callback[j](r.hash, r.blob); err != nil {
-                               log.Fatalf("Callback returns error: %d %v", i, err)
+                       for j := range callback {
+                               if err := callback[j](r.hash, r.blob); err != nil {
+                                       log.Fatalf("Callback returns error: %d %v", i, err)
+                               }
                        }
                }
        }
```

That's it: a single task per goroutine, a single task may generate either one or two transactions and send it/them to the corresponding signer job. A signer job can sign either one or two transactions depending on the result channel size.

Please, try to embed new code into the old one as much as possible. It allows to keep the codebase small and readable, and this way you don't need a lot of time to understand parts of the system. Also, small diff makes it easier to review and merge.

Makefile
```diff
@@ -30,6 +30,7 @@ help:
 	start.GoFourNodes10wrk start.GoFourNodes30wrk start.GoFourNodes100wrk \
 	start.GoFourNodes25rate start.GoFourNodes50rate start.GoFourNodes60rate start.GoFourNodes300rate start.GoFourNodes1000rate \
 	start.GoFourNodes4RPC10wrk start.GoFourNodes4RPC30wrk start.GoFourNodes4RPC100wrk \
+	start.Go4CN4RPC10wrk start.Go4CN4RPC30wrk start.Go4CN4RPC100wrk \
```
AnnaShaleva: This name is very similar to `start.GoFourNodes4RPC10wrk`, it's not possible to distinguish these two cases being out of the context (and in two months I'll definitely be out of the context of this task). Let's follow the existing naming and just add `Conflicts` suffix to it, something like `start.GoFourNodes4RPCConflicts10wrk`. Suggestions are welcomed.

Makefile
```diff
@@ -217,6 +218,16 @@ start.GoFourNodes4RPC30wrk:
 start.GoFourNodes4RPC100wrk:
 	./runner.sh --rpc go4 -d "Go4x4" -m wrk -w 100 -z 5m -t 30s
 
+#	## Conflicting transaction:
```
AnnaShaleva: s/transaction/transactions

cmd/internal/conflict.go
```diff
@@ -0,0 +1,133 @@
...
+func pickRandomNodePair(n int) (int, int) {
+	a := rand.IntN(n)
+	b := rand.IntN(n)
+	for b == a {
+		b = rand.IntN(n)
+	}
+	return a, b
+}
+
+func (d *doer) SendConflictPairs(ctx context.Context) {
```
AnnaShaleva: Ideally, I'd like to avoid `doer` interface extension. We already have `doer.Sender`, I think it's enough. We may add an additional parameter to this method, something like this:
```
anna@kiwi:~/Documents/GitProjects/nspcc-dev/neo-bench$ git diff
diff --git a/cmd/internal/worker.go b/cmd/internal/worker.go
index 34b450f..d23740c 100644
--- a/cmd/internal/worker.go
+++ b/cmd/internal/worker.go
@@ -215,8 +215,15 @@ func NewWorkers(opts ...WorkerOption) (Worker, error) {
        return w, nil
 }
 
+type mode byte
+
+const (
+       single mode = 1 + iota
+       pair
+)
+
 // idx defines the order of the transaction being sent and can be more than overall transactions count, because retransmission is supported.
-func (d *doer) worker(ctx context.Context, idx *atomic.Int64, start time.Time) {
+func (d *doer) worker(ctx context.Context, typ mode, idx *atomic.Int64, start time.Time) {
        var (
                done           = ctx.Done()
                timer          = time.NewTimer(d.timeLimit)
@@ -235,26 +242,41 @@ loop:
                        if d.dump.TransactionsQueue.Len() == 0 {
                                return
                        }
-                       tx, err := d.dump.TransactionsQueue.Get()
-                       if err != nil {
-                               log.Fatalf("cannot dequeue transaction: %s", err)
-                               return
+                       var tx, conflict string
+                       switch typ {
+                       case single:
+                               blob, err := d.dump.TransactionsQueue.Get()
+                               if err != nil {
+                                       log.Fatalf("cannot dequeue transaction: %s", err)
+                                       return
+                               }
+                               tx = blob.(string)
+                       case pair:
+                               d.dumpLock.Lock()
+                               blob1, err := d.dump.TransactionsQueue.Get()
+                               if err != nil {
+                                       log.Fatalf("cannot dequeue transaction: %s", err)
+                                       return
+                               }
+                               blob2, err := d.dump.TransactionsQueue.Get()
+                               if err != nil {
+                                       log.Fatalf("cannot dequeue transaction: %s", err)
+                                       return
+                               }
+                               d.dumpLock.Unlock
+                               tx = blob1.(string)
+                               conflict = blob2.(string)
                        }
-                       if err := d.cli.SendTX(ctx, tx.(string)); err != nil {
-                               if errors.Is(err, ErrMempoolOOM) {
-                                       err := d.dump.TransactionsQueue.Put(tx.(string))
-                                       if err != nil {
-                                               log.Printf("failed to re-enqueue transaction: %s\n", err)
-                                               d.countErr.Add(1)
+
+                       var wg sync.WaitGroup
+                       for range typ {
+                               wg.Go(func() {
+                                       if err := d.cli.SendTX(ctx, tx.(string)); err != nil {
+                                               // ...
                                        }
-                                       time.Sleep(d.mempoolOOMDelay)
-                               } else {
-                                       d.countErr.Add(1)
-                               }
-                               continue loop
-                               // d.stop()
-                               // return
+                               })
                        }
+                       wg.Wait()
 
                        since := time.Since(start)
                        count := d.countTxs.Add(1)
anna@kiwi:~/Documents/GitProjects/nspcc-dev/neo-bench$ 
```

cmd/internal/conflict.go
```diff
@@ -0,0 +1,133 @@
...
+					return
+				default:
+				}
+
+				blobA, blobB, ok := d.claimConflictPair()
+				if !ok {
+					return
+				}
+
+				nodeA, nodeB := pickRandomNodePair(nodeCount)
```
AnnaShaleva: This helper is likely unavoidable, so let's keep it and see how it goes with the refactored `Sender`. Along with `SendTXToNode`.

AnnaShaleva (commented):
There's still a lot of complication and I'd like the code to be more simple and unified. Let's refactor Sender and generator and see the result.

cmd/internal/client.go
Turalchik: Checked, and there doesn't seem to be a dedicated neorpc error for this.

cmd/internal/conflict.go
Turalchik: Feels like we just squashed the two paths into one worker with a switch `typ` re-evaluated every loop iteration.

cmd/internal/client.go
```diff
@@ -118,16 +121,32 @@ func (c *RPCClient) GetVersion(ctx context.Context) (*result.Version, error) {
 
 // SendTX sends transaction.
 func (c *RPCClient) SendTX(ctx context.Context, tx string) error {
+	idx := int(c.inc.Add(1) % c.len)
+	return c.sendTXToAddr(ctx, tx, idx)
+}
+
+func (c *RPCClient) SendTXToNode(ctx context.Context, tx string, nodeIdx int) error {
```
AnnaShaleva, line 130: Exported method should have a comment.

cmd/internal/client.go
AnnaShaleva: OK, at least then let's use `core.ErrHasConflicts`'s text instead of the raw string.

cmd/internal/client.go
```diff
@@ -163,8 +182,16 @@ func (c *RPCClient) GetBlockCount(ctx context.Context) (int, error) {
...
+func (c *RPCClient) doRPCCall(ctx context.Context, call string, result any, client *fasthttp.Client) error {
+	idx := int(c.inc.Add(1) % c.len)
+	return c.doRPCCallToAddr(ctx, call, result, client, idx)
+}
+
+func (c *RPCClient) doRPCCallToAddr(_ context.Context, call string, result any, client *fasthttp.Client, idx int) error {
+	idx %= int(c.len)
+	if idx < 0 {
+		idx += int(c.len)
+	}
```
AnnaShaleva: How is this possible? Don't complicate code for no reason.

cmd/internal/client.go
```diff
@@ -163,8 +182,16 @@ func (c *RPCClient) GetBlockCount(ctx context.Context) (int, error) {
...
 
-func (c *RPCClient) doRPCCall(_ context.Context, call string, result any, client *fasthttp.Client) error {
-	idx := c.inc.Add(1) % c.len
+func (c *RPCClient) doRPCCall(ctx context.Context, call string, result any, client *fasthttp.Client) error {
+	idx := int(c.inc.Add(1) % c.len)
+	return c.doRPCCallToAddr(ctx, call, result, client, idx)
+}
+
+func (c *RPCClient) doRPCCallToAddr(_ context.Context, call string, result any, client *fasthttp.Client, idx int) error {
+	idx %= int(c.len)
```
AnnaShaleva: All callers of `doRPCCallToAddr` use a proper index. So what is the purpose of this additional check? You may add a comment to `doRPCCallToAddr` about `idx` constraints.

cmd/internal/gen.go
```diff
@@ -52,6 +53,8 @@ const (
 	GASTransfer = "gas"
 	// ContractTransfer is the type of deployed NEP17 contract transfer tx.
 	ContractTransfer = "nep17"
+	// ConflictTransfer is the type of conflicting transaction pairs.
```
AnnaShaleva: s/pairs/pair

cmd/internal/gen.go
```diff
@@ -40,8 +40,9 @@ type (
 	}
 
 	txRequest struct {
-		tx  *transaction.Transaction
-		acc *wallet.Account
+		tx       *transaction.Transaction
+		acc      *wallet.Account
+		conflict *transaction.Transaction
```
AnnaShaleva, line 46: Comment.

cmd/internal/gen.go
```diff
@@ -77,14 +80,17 @@ func newTransferTx(p *keys.PrivateKey, contractHash, toAddr util.Uint160) *trans
 		panic(w.Err)
 	}
 
-	script := w.Bytes()
+	return newTx(p, w.Bytes(), transaction.CalledByEntry)
+}
+
+func newTx(p *keys.PrivateKey, script []byte, scopes transaction.WitnessScope) *transaction.Transaction {
```
AnnaShaleva: Don't need `scopes` parameter. All of them can use CalledByEntry.

AnnaShaleva: Replace private key parameter with scripthash.

cmd/internal/gen.go
```diff
@@ -196,6 +220,7 @@ func genTxWorker(n int, ch <-chan txRequest, out chan<- txBlob) {
 	for tr := range ch {
 		tx := *tr.tx
 		tx.Nonce = uint32(baseNonce | i)
+		i++
```
AnnaShaleva: Move it right after `out <- txBlob{` sending -- it's what we're accumulating with `i` counter.

cmd/internal/gen.go
```diff
@@ -213,6 +238,32 @@ func genTxWorker(n int, ch <-chan txRequest, out chan<- txBlob) {
...
+			continue
+		}
+
+		conflict := *tr.conflict
+		conflict.Nonce = uint32(baseNonce | i)
 		i++
+		conflict.Attributes = []transaction.Attribute{{
+			Type:  transaction.ConflictsT,
+			Value: &transaction.Conflicts{Hash: tx.Hash()},
+		}}
```
AnnaShaleva: This should be done by `Generate`; `genTxWorker`'s responsibility is to sign the generated transaction.

cmd/internal/gen.go
```diff
@@ -213,6 +238,32 @@ func genTxWorker(n int, ch <-chan txRequest, out chan<- txBlob) {
 			blob: base64.StdEncoding.EncodeToString(buf.Bytes()),
 		}
 
+		if tr.conflict == nil {
+			continue
+		}
+
+		conflict := *tr.conflict
```
AnnaShaleva: A large duplicating code fragment. Move to a separate function, reuse.

cmd/internal/reader.go
```diff
@@ -42,6 +43,10 @@ func ReadDump(from string) *Dump {
 	dump.BenchOptions.DecodeBinary(rd)
 
 	count := dump.BenchOptions.TxCount
+	if strings.ToLower(dump.BenchOptions.TransferType) == ConflictTransfer {
+		// TxCount is the number of conflicting pairs, i.e. 2 dumped transactions each.
```
AnnaShaleva: `TxCount`?

AnnaShaleva: Remove `, i.e. 2 dumped transactions each.`.

cmd/internal/worker.go
```diff
@@ -215,23 +216,38 @@ func NewWorkers(opts ...WorkerOption) (Worker, error) {
...
 		done           = ctx.Done()
 		timer          = time.NewTimer(d.timeLimit)
 		localTxCounter int64
+		nodeCount      int
 	)
+	if typ == pair {
+		nodeCount = d.cli.AddrCount()
+	}
 
-loop:
```
AnnaShaleva, line 226: Let's keep it, for better readability.

cmd/internal/worker.go
```diff
@@ -240,35 +256,91 @@ loop:
...
+				defer wg.Done()
+				d.handleSendResult(conflict, d.cli.SendTXToNode(ctx, conflict, nodeB), start)
+			}()
+			wg.Wait()
+		}
+	}
+}
+
+// dequeuePair atomically dequeues both halves of a conflicting pair so that
+// concurrent workers can't split a pair across each other.
```
AnnaShaleva: s/both halves of a conflicting pair/a pair of transactions

cmd/internal/worker.go
```diff
@@ -240,35 +256,91 @@ loop:
...
+func (d *doer) dequeuePair() (tx, conflict string, ok bool) {
```
AnnaShaleva: No named return, please.

cmd/internal/worker.go
```diff
@@ -240,35 +256,91 @@ loop:
...
+}
+
+func pickRandomNodePair(n int) (int, int) {
+	a := rand.IntN(n)
+	b := rand.IntN(n)
+	for b == a {
+		b = rand.IntN(n)
+	}
+	return a, b
+}
```
AnnaShaleva: NACK. Just use the same approach as for a single receiver. Gives more deterministic result.

cmd/internal/worker.go
```diff
@@ -240,35 +256,91 @@ loop:
...
+			}
+
+			nodeA, nodeB := pickRandomNodePair(nodeCount)
+
+			var wg sync.WaitGroup
+			wg.Add(2)
+			go func() {
+				defer wg.Done()
+				d.handleSendResult(tx, d.cli.SendTXToNode(ctx, tx, nodeA), start)
+			}()
```
AnnaShaleva: Learn Go specification, read Go release notes on every release, keep your language knowledge up-to-date.

`wg.Go(...)`.

cmd/internal/worker.go
```diff
@@ -240,35 +256,91 @@ loop:
...
-			count := d.countTxs.Add(1)
 			localTxCounter++
-			d.rpsReporter(float64(count) / since.Seconds())
-
 			if d.threshold > 0 {
 				waitFor := time.Until(start.Add(time.Duration(d.threshold.Nanoseconds() * (localTxCounter + 1))))
 				if waitFor > 0 {
 					time.Sleep(waitFor)
 				}
 			}
```
AnnaShaleva: I suppose the same delay should be applied to a conflict pair. Check it and implement if so.

cmd/internal/worker.go
```diff
@@ -215,23 +216,38 @@ func NewWorkers(opts ...WorkerOption) (Worker, error) {
 	return w, nil
 }
 
+type txMode byte
+
+const (
+	single txMode = 1 + iota
```
AnnaShaleva: It was just an example. Use appropriate names, `internal.single` doesn't tell me anything.

cmd/internal/worker.go
```diff
@@ -395,18 +467,30 @@ func (d *doer) parse(ctx context.Context, startBlock int, lastTime *uint64) (las
...
+			d.worker(ctx, typ, idx, start)
 		})
 	}
 
 	d.waiter.Wait()
 
+	d.reportSendResult(start)
+}
+
+func (d *doer) reportSendResult(start time.Time) {
```
AnnaShaleva: Useless change.

cmd/internal/client.go
```diff
@@ -44,6 +44,9 @@ const DefaultTimeout = time.Second * 30
 var (
 	// ErrMempoolOOM is returned from `sendrawtransaction` when node cannot process transaction due to mempool OOM.
 	ErrMempoolOOM = errors.New("node cannot process transaction due to mempool OOM")
+	// ErrConflict is returned from `sendrawtransaction` when the node rejects a
+	// transaction because it conflicts via the Conflicts attribute.
```
AnnaShaleva: s/conflicts/conflicts with mempool or chain content

AnnaShaleva (commented):
Less changes, much cleaner.

cmd/internal/reader.go
Turalchik: Yes, because https://github.com/nspcc-dev/neo-bench/pull/224#discussion_r3956179932.

cmd/internal/gen.go
Turalchik: Moved nonce assignment into `Generate` too, since the `Conflicts` attribute needs the other tx's hash, which depends on its nonce.

cmd/internal/worker.go
```diff
@@ -231,7 +238,12 @@ loop:
 		case <-timer.C:
 			return
 		default:
-			idx.Add(1)
+		}
+
+		idx.Add(1)
```
AnnaShaleva: Useless change.

cmd/internal/client.go
```diff
@@ -118,16 +122,42 @@ func (c *RPCClient) GetVersion(ctx context.Context) (*result.Version, error) {
...
+	return c.sendTXToAddr(ctx, tx, idx)
+}
+
+// SendTXToNode sends transaction to the node at nodeIdx.
+func (c *RPCClient) SendTXToNode(ctx context.Context, tx string, nodeIdx int) error {
+	return c.sendTXToAddr(ctx, tx, nodeIdx)
+}
+
+// AddrCount returns the number of configured RPC addresses.
+func (c *RPCClient) AddrCount() int {
```
AnnaShaleva: Remove this method, it's not needed. Use `c.len` directly.

cmd/internal/client.go
```diff
@@ -118,16 +122,42 @@ func (c *RPCClient) GetVersion(ctx context.Context) (*result.Version, error) {
...
+
+// PickNodePair returns two distinct node indices.
+func (c *RPCClient) PickNodePair() (int, int) {
+	base := c.inc.Add(2)
+	a := int(base % c.len)
+	b := int((base + 1) % c.len)
+	return a, b
+}
+
+func (c *RPCClient) sendTXToAddr(ctx context.Context, tx string, idx int) error {
```
AnnaShaleva: Remove this method. `SendTxToNode` is sufficient for our purposes. `SendTX` can reuse `SendTxToNode` directly.

cmd/internal/gen.go
```diff
@@ -188,31 +234,30 @@ func Generate(ctx context.Context, opts BenchOptions, callback ...GenerateCallba
 	return &dump
 }
 
-func genTxWorker(n int, ch <-chan txRequest, out chan<- txBlob) {
-	baseNonce := n << 24 // 255 possible workers and 16M transactions should be enough
-	i := 0
-
+func genTxWorker(ch <-chan txRequest, out chan<- txBlob) {
```
AnnaShaleva, line 236: Don't remove `n` from it, otherwise it's too easy to get nonce conflicts. Revert this change.

cmd/internal/reader.go
```diff
@@ -42,6 +43,10 @@ func ReadDump(from string) *Dump {
 	dump.BenchOptions.DecodeBinary(rd)
 
 	count := dump.BenchOptions.TxCount
+	if strings.ToLower(dump.BenchOptions.TransferType) == ConflictTransfer {
+		// TxCount is the number of conflicting pairs.
```
AnnaShaleva: s/TxCount/count

cmd/internal/worker.go
```diff
@@ -240,35 +252,87 @@ loop:
...
+			nodeA, nodeB := d.cli.PickNodePair()
 
-			if d.threshold > 0 {
-				waitFor := time.Until(start.Add(time.Duration(d.threshold.Nanoseconds() * (localTxCounter + 1))))
-				if waitFor > 0 {
-					time.Sleep(waitFor)
-				}
+			var wg sync.WaitGroup
+			wg.Go(func() {
+				d.handleSendResult(tx, d.cli.SendTXToNode(ctx, tx, nodeA), start)
```
AnnaShaleva: The result is unhandled. Move throttling inside routines.

cmd/internal/gen.go
AnnaShaleva: OK, need to think about it, we still need global nonce.

AnnaShaleva:
Need to rebase onto master.

cmd/internal/gen.go
AnnaShaleva: OK, I see, we need nonce to be determined before the conflicting transaction generation. So let it be, but we still need a `baseNonce` to be presented at the `Generate` level. Use `baseNonce` as a base and add some offest to it, otherwise it's too easy to get a nonce collision.

Turalchik: What is meant by “nonce collision”?

AnnaShaleva: Two transactions with identical content (same transfer details, in our case) must have different nonce, otherwise hashes are equal.

AnnaShaleva: Use the same approach as it's done on master branch, just move nonce-related code to the `Generate` level.

cmd/internal/gen.go
```diff
@@ -155,7 +179,26 @@ func Generate(ctx context.Context, opts BenchOptions, callback ...GenerateCallba
...
+				conflict := *r.conflict
+				conflict.Nonce = baseNonce | uint32(2*i+1)
+				conflict.Attributes = []transaction.Attribute{{
+					Type:  transaction.ConflictsT,
+					Value: &transaction.Conflicts{Hash: tx.Hash()},
+				}}
+				task.conflict = &conflict
+			}
+
+			txCh[i%len(txCh)] <- task
```
AnnaShaleva:
```suggestion
			txCh[wrkIdx] <- task
```

cmd/internal/gen.go
```diff
@@ -188,31 +233,30 @@ func Generate(ctx context.Context, opts BenchOptions, callback ...GenerateCallba
...
 
-		buf.Reset()
-		tx.EncodeBinary(buf.BinWriter)
-
-		if buf.Err != nil {
-			log.Fatalf("Could not prepare transaction: %d %v", i, buf.Err)
-		}
+func signAndEncode(buf *io.BufBinWriter, acc *wallet.Account, tx *transaction.Transaction) txBlob {
+	if err := acc.SignTx(netmode.PrivNet, tx); err != nil {
+		log.Fatalf("Could not sign tx: %v", err)
```
AnnaShaleva, line 249: %w

cmd/internal/gen.go
```diff
@@ -188,31 +233,30 @@ func Generate(ctx context.Context, opts BenchOptions, callback ...GenerateCallba
...
+	}
 
-		out <- txBlob{
-			hash: tx.Hash().String(),
-			blob: base64.StdEncoding.EncodeToString(buf.Bytes()),
-		}
+	buf.Reset()
+	tx.EncodeBinary(buf.BinWriter)
+	if buf.Err != nil {
+		log.Fatalf("Could not prepare transaction: %v", buf.Err)
```
AnnaShaleva, line 255: %w while we're here.

cmd/internal/gen.go
```diff
@@ -52,6 +54,8 @@ const (
 	GASTransfer = "gas"
 	// ContractTransfer is the type of deployed NEP17 contract transfer tx.
 	ContractTransfer = "nep17"
+	// ConflictTransfer is the type of conflicting transaction pair.
+	ConflictTransfer = "conflict"
```
AnnaShaleva, line 58: Looks good now. We need an additional `make gen` target to generate those conflicts, like it's done with other transaction files. This file should be read by `start.GoFourNodes4RPCConflicts10wrk` and etc.

AnnaShaleva (commented):
Run the final version from the clear environment to check everything (along with transaction generation) works fine in the conflicts scenario.

cmd/internal/gen.go
Turalchik: > log.Fatalf does not support error-wrapping directive

Turalchik:
I ran this target `start.GoFourNodes4RPCConflicts10wrk`, and at the same time I ran the script from this pull request https://github.com/nspcc-dev/neo-go/pull/4394, and I got the [result](https://github.com/user-attachments/files/32015521/mempool.log) of the script’s execution and the [statistics](https://github.com/user-attachments/files/32015494/Go4x4Conflict__10.log) for the run itself.

cmd/internal/gen.go
```diff
@@ -155,7 +179,26 @@ func Generate(ctx context.Context, opts BenchOptions, callback ...GenerateCallba
 				log.Fatal(ctx.Err())
 			}
 
-			txCh[i%len(txCh)] <- txR[i%len(txR)]
+			wrkIdx := uint32(i % len(txCh))
+			baseNonce := wrkIdx << 24
```
AnnaShaleva: When moving code to a new location, always preserve the original comments.

Makefile
```diff
@@ -120,6 +122,7 @@ pull:
 gen: $(BUILD_DIR)/dump.NEO.$(NEOBENCH_FROM_COUNT).$(NEOBENCH_TO_COUNT).txs
 gen: $(BUILD_DIR)/dump.GAS.$(NEOBENCH_FROM_COUNT).$(NEOBENCH_TO_COUNT).txs
 gen: $(BUILD_DIR)/dump.NEP17.$(NEOBENCH_FROM_COUNT).$(NEOBENCH_TO_COUNT).txs
+gen: $(BUILD_DIR)/dump.conflict.$(NEOBENCH_FROM_COUNT).$(NEOBENCH_TO_COUNT).txs
```
AnnaShaleva, line 125: Not sure if this file is used, let me check.

AnnaShaleva (commented):
Otherwise LGTM.

Makefile
AnnaShaleva: Yes, it's used as a docker volume:
https://github.com/nspcc-dev/neo-bench/blob/0cc4a9aac77cf4718c920070c9ae1ddb275b2831/.docker/rpc/docker-compose.bench.yml#L7

So I need you to search for usages of `NEOBENCH_TYPE` variable and adjust the corresponding documentation: README, `runner.sh` manual text and etc. We added additional value to `--benchmark` flag, it should be reflected across the doc.

AnnaShaleva (approved):
Good job.

